PR #72, opened 2022-06-09
add doc for Inv()

shanzhaowang99:
Added the doc of Inv()


aerogjy (approved):
I add a note directly to address the mathematical concept of Inv for Lie Group/Lie Algebra.

PR #88, opened 2022-07-03
Feature/optim

wang-chen:
Add Gauss-Newton Optimizer and Robust Kernel function such as Huber.


pypose/optim/optimizer.py
```diff
@@ -1,39 +1,244 @@
...
+            return self.model(*inputs)
+        else:
+            return self.model(inputs)
+
+    def residual(self, outputs, targets):
+        return outputs if targets is None else outputs - targets
+
+    def kernel_forward(self, module, inputs, outputs):
+        # eps is to prevent grad of sqrt() from being inf
+        eps = finfo(outputs.dtype).eps
```
huyaoyu, line 49: This raises an exception if outputs.dtype is an integer type.

pypose/optim/optimizer.py
```diff
@@ -1,39 +1,244 @@
...
+            Early Stopping with error: 5.21540641784668e-07
+        '''
+        R = self.model(inputs, targets)
+        for pg in self.param_groups:
+            numels = [p.numel() for p in pg['params'] if p.requires_grad]
+            J = modjac(self.model, inputs=(inputs, targets), flatten=True)
+            R, J = self.corrector(R = R, J = J)
+            D = self.solver(A = J, b = -R.view(-1, 1)).split(numels)
+            [p.add_(d.view(p.shape)) for p, d in zip(pg['params'], D) if p.requires_grad]
+        return self.model.loss(inputs, targets)
```
huyaoyu, line 209: self.model is required to have the loss() function. So it is not a simple nn.module.
The same comment for the LM.

wang-chen: the loss() function is from the standardized model, not the model that user provide any more: see this line

https://github.com/pypose/pypose/blob/cca97237eb0120e31d8acf50097b70a8a400a0ef/pypose/optim/optimizer.py#L141

wang-chen: I could change the `loss` method to `_loss` if you think this name may conflict with the users' self-defined method.

PR #95, opened 2022-07-21
Yuheng cov pro

haleqiu:


wang-chen:
Can you add description for what you have done in this pull request?

haleqiu:
Sure. This pr considers the inputted noises as a vector instead of a constant number. Fixed bugs in the cov propagation. Fixed the loss function when supervising the rotation in the `imu_correction.py`. Also I detached the input tensor of the `cov_propagation` to stop the gradient propagation.

wang-chen:
I cannot reproduce the results of IMU preintegration. The generated figure show no covariance.

pypose/module/imu_preintegrator.py
```diff
@@ -66,11 +79,21 @@ def forward(self, dt, gyro, acc, rot:pp.SO3=None, gyro_cov=None, acc_cov=None, i
...
+              :code:`pos`: initial position, :code:`rot`: initial rotation, :code:`vel`: initial
+              velocity, with the shape :math:`(B, H_{in})`
+
+            - output: The output is a :obj:`dict` of integration state.
+              It contains the ``pos``: position, ``rot``: rotation, ``vel``: velocity.
+              Each output has the shape :math:`(B, F, H_{out})`, where :math:`H_{out}` is
+              the corresponding signal dimension. If the flag :obj:`prop_cov` is ``True``, the
+              output will include ``cov``, the covariance matrix with shape :math:`(B, 9, 9)`.
 
         IMU Measurements Integration:
```
wang-chen, line 99: what is `b_i^g` `b_i^a` and `b_i^a`? better to add an explanation.

pypose/module/imu_preintegrator.py
```diff
@@ -8,34 +8,47 @@ class IMUPreintegrator(nn.Module):
...
+        
         # Initial status of IMU: (pos)ition, (rot)ation, (vel)ocity, (cov)ariance
         self.register_buffer('gravity', torch.tensor([0, 0, gravity]), persistent=False)
         self.register_buffer('pos', self._check(pos).clone(), persistent=False)
         self.register_buffer('rot', self._check(rot).clone(), persistent=False)
         self.register_buffer('vel', self._check(vel).clone(), persistent=False)
         self.register_buffer('cov', torch.zeros(1, 9, 9), persistent=False)
+        self.register_buffer('gyro_cov', gyro_cov, persistent=False)
+        self.register_buffer('acc_cov', acc_cov, persistent=False)
+        self.Rij = None # th  referenced rotation of the covariance
```
wang-chen, line 51: comments of this line are unclear.

pypose/module/imu_preintegrator.py
```diff
@@ -12,30 +12,43 @@ class IMUPreintegrator(nn.Module):
...
         vel (torch.Tensor, optional): initial position. Default: torch.zeros(3)
         gravity (float, optional): the gravity acceleration. Default: 9.81007
-        gyro_cov (float, optional): covariance of the gyroscope. Default: (1.6968e-4)**2
-        acc_cov (float, optional): covariance of the accelerator. Default: (2e-3)**2
+        gyro_cov: covariance of the gyroscope. If the covariance of the x, y and z axises are
+            different, the :obj:`gyro_cov` use the form :code:`torch.Tensor(gyro_cov_x, gyro_cov_y,
+            gyro_cov_z)`. Default: (3.2e-3)**2
+        acc_cov: covariance of the accelerator. If the covariance of the x, y and z axises are
+            different, the :obj:`acc_cov` use the form :code:`torch.Tensor(acc_cov_x, acc_cov_y,
+            acc_cov_z)`. Default: (8e-2)**2
```
wang-chen, line 20: Suggest to change: covariance of the gyroscope. Default: (3.2e-3)**2. Please use the form of :code:`torch.Tensor(gyro_cov_x, gyro_cov_y, gyro_cov_z)`, if the covariance on the three axes is different.

pypose/module/imu_preintegrator.py
```diff
@@ -277,8 +319,8 @@ def integrate(self, init_state, dt, gyro, acc, rot:pp.SO3=None):
         incre_t = torch.cumsum(dt, dim = 1)
         incre_t = torch.cat([torch.zeros(B, 1, 1, dtype=dt.dtype, device=dt.device), incre_t], dim =1)
 
-        return {'acc':acc, 'vel':incre_v[...,1:,:], 'pos':incre_p[:,1:,:], 'rot':incre_r[:,1:,:],
-                't':incre_t[...,1:,:], 'dr': dr[:,1:,:], 'dt': dt}
+        return {'acc':a, 'vel':incre_v[...,1:,:], 'pos':incre_p[:,1:,:], 'rot':incre_r[:,1:,:],
```
wang-chen, line 322: Do you need to change the doc by changing the `acc` to `a`?

pypose/module/imu_preintegrator.py
wang-chen: Do you have to add `self.Rij` here?

wang-chen (changes requested):

pypose/module/imu_preintegrator.py
haleqiu: In our current implementation, `b_i^g` is not modeled in the `integrator`. The input to the integrator shall be "de-bias". I will suggest to remove this notation.

pypose/module/imu_preintegrator.py
haleqiu: Yes, with the flag of `self.reset`, if you want to track the correct covariance with a non-zero `init_cov`, you need to maintain both the `self.cov` and `self.Rij`.

pypose/module/imu_preintegrator.py
haleqiu: That's a good suggestion, I just updated.

pypose/module/imu_preintegrator.py
haleqiu: In order to show the difference between raw acceleration `acc` and "gravity-corrected" acceleration `a`. I will change the key to `a`. In our documentation, all `acc` denote the "gravity-corrected" acceleration, there are no misleading, so I think we don't need to update the doc.

pypose/module/imu_preintegrator.py
haleqiu: I just changed my mind, maybe we will need to model the bias in our future. I will add the explanation.

haleqiu:
@wang-chen Could you please review the changes?

wang-chen (approved):

PR #98, opened 2022-07-22
add pytest with circleci

wang-chen:


test/lietensor/test_lietensor.py
```diff
@@ -0,0 +1,297 @@
...
+        outputs = net(inputs)
+        loss = outputs.abs().sum()
+        loss.backward()
+        optimizer.step()
+        scheduler.step()
+        print(loss)
+    print(net.weight)
+    print("Parameter:", count_parameters(net))
+
+    SO3I = pp.identity_SO3(1, 3, device=device, dtype=torch.float64)
```
Xanthorapedia, line 62: It may be more readable and scalable to use [@pytest.mark.parametrize](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) to generate a cartesian product of test inputs (`identity_<ltype>` x shape in this case).

test/lietensor/test_lietensor.py
```diff
@@ -0,0 +1,297 @@
...
+    a = pp.randn_se3(8, dtype=torch.double)
+    b = X.Adj(a)
+    assert (b.Exp() * X - X * a.Exp()).abs().mean() < 1e-7
+    J = X.Jinvp(a)
+    print(J)
+
+    X = pp.randn_SE3(6, requires_grad=True)
+    a = pp.randn_se3(6)
+    b = X.AdjT(a)
+    assert (X * b.Exp() - a.Exp() * X).abs().mean() < 1e-6
```
Xanthorapedia, line 127: A `test_*` function should not include more than one assertion. Otherwise, subsequent tests will not be run if an earlier check fails.

Xanthorapedia (changes requested):
I would suggest the following major changes:
1. Parametrize tests to separate test input/output from test code. This makes it much easier and less error-prone to add more test cases.
2. Make tests regressive by fixing random seeds and hardcoding expected outputs.
3. Split `test_*` functions with multiple assertions.
4. Use assertion functions provided by torch/numpy/pytest instead of bare `assert` for better error reporting.

test/lietensor/test_lietensor.py
huyaoyu: I think right now the guideline is not so clear. Maybe we need some professionals to give us some hints on how to come up with good tests. Or the structure for testing large-scale projects.

I have multiple assertions in all my test functions. However, a single test function only focuses on a group of related features. If one assertion fails, it probably means other related assertions may also fail. And it often happens that later assertions are made on results that depends on the previous, failed assertions. If the previous fails, the latter will probably fail. And I strongly recommend that all test scripts should be able to run by themselves. This allows us to only focus on a failed test script, iterating on it without the burden of running all the tests every time.

Xanthorapedia: Agreed we should have a guideline. 

> If the previous fails, the latter will probably fail.

I see. If tests are strongly dependent then this is ok. I was trying to say that ideally, we may want each test to only cover one property of the output or even a small piece of code, hence the word "unit". It helps us pin down the problem faster if we know which lines of code cause the problem.

> And I strongly recommend that all test scripts should be able to run by themselves.

Fortunately, pytest has a [`-k` option](https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name) for filtering tests by name.

examples/module/imu/readme.md
```diff
@@ -29,7 +29,9 @@ An example for IMU Preintegration
 
 ## Run:
 
-        python imu_preint.py --dataroot DATAROOT --datadrive 0018 0022
+        python imu_integrator.py --dataroot DATAROOT --datadrive 0018 0022
+
+Note: change DATAROOT to the folder you select.
```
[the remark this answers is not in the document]
wang-chen: done

test/lietensor/test_lietensor.py
```diff
@@ -0,0 +1,297 @@
...
+    x = pp.so3(a)
+    y = pp.randn_so3(3, sigma=0.1, dtype=torch.float64, requires_grad=True, device=device)
+    a = pp.randn_se3(3, sigma=0.1, requires_grad=True)
+    b = pp.SE3_type.randn(3, sigma=0.1, requires_grad=True)
+
+    assert y.is_leaf and x.is_leaf and a.is_leaf and b.is_leaf
+
+    (pp.Log(y.Exp())**2).sin().sum().backward()
+
+    print(y)
```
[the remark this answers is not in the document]
wang-chen: done

test/lietensor/test_lietensor.py
```diff
@@ -0,0 +1,297 @@
...
+
+    r = pp.Retr(I, a)
+    p = I.Act(a)
+
+    print(r, p)
+
+    X = pp.randn_SE3(8, requires_grad=True, dtype=torch.double)
+    a = pp.randn_se3(8, dtype=torch.double)
+    b = X.Adj(a)
+    assert (b.Exp() * X - X * a.Exp()).abs().mean() < 1e-7
```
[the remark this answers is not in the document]
wang-chen: done

test/lietensor/test_lietensor.py
wang-chen: This is to test whether the grammar works. not works for specific data.

test/lietensor/test_lietensor.py
```diff
@@ -0,0 +1,297 @@
...
+from torch import nn
+import pypose.optim.solver as ppos
+import pypose.optim.kernel as ppok
+import pypose.optim.corrector as ppoc
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+
+def test_lietensor():
+
+    a = torch.randn(3,3)*0.1
```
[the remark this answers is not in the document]
wang-chen: not yet, but it will not fixed in the pr for pytest

Xanthorapedia (approved):

PR #99, opened 2022-07-25
Dmpc nonlinear dynamics

apandey02:
[This is a test pull request.]


pypose/module/dynamics.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def set_linearization_point(self, state, input):
+        self.state, self.input = state, input
+
+    @property
+    def A(self):
+        if hasattr(self, '_A'):
+            return self._A
+        else:
+            func = lambda x: self.state_transition(x, self.input)            # DH: TYPO.  So has the Jacobian been ever tested yet??
```
smanist, line 62: Remove my comment

pypose/module/dynamics.py
```diff
@@ -0,0 +1,219 @@
+from numpy import vectorize
+import torch as torch
+import torch.nn as nn
+from torch.autograd.functional import jacobian
+
+class _System(nn.Module):                                                # DH: Please follow the documentation style in LTI to document this class.
```
smanist, line 6: Does the doc compile?

pypose/module/dynamics.py
```diff
@@ -0,0 +1,219 @@
...
+            return jacobian(func, self.state, **self.jacargs)
+ 
+    @property
+    def D(self):
+        if hasattr(self, '_D'):
+            return self._D
+        else:
+            func = lambda x: self.observation(self.state, x)
+            return jacobian(func, self.input, **self.jacargs)
+
```
smanist, line 88: I think we should define properties c1 and c2, as the constant bias from linearization (when the linearization point is not at equilibrium).  See how c1 and c2 are used in LTI.

test/nonlinear_dyn_cartpole.py
```diff
@@ -0,0 +1,92 @@
...
+
+    ### Jacobian computations
+    # Find jacobinas at 1000th step
+    jacob_state, jacob_input = state_all[999,:].T, input[999]
+    cartPoleSolver.set_linearization_point(jacob_state,jacob_input)
+    A = cartPoleSolver.A
+    B = cartPoleSolver.B
+    C = cartPoleSolver.C
+    D = cartPoleSolver.D
+    print(A)
```
smanist, line 89: How do you verify the numerical correctness of the Jacobian?

test/nonlinear_dyn_lorenz
```diff
@@ -0,0 +1,76 @@
+import pypose.module.dynamics as ppmd
```
smanist, line 1: First of all, the file name....

test/nonlinear_dyn_lorenz
```diff
@@ -0,0 +1,76 @@
...
+    sol = torch.tensor(sol.y).float()
+    
+    lorenzSolver = LorenzAttractor(t_step)
+
+    pred = torch.zeros(np.shape(sol))
+    state = torch.tensor(IC).float()
+    pred[:,0] = state;
+
+    stateTransitionOptim = torch.optim.Adam(lorenzSolver.parameters(),lr=0.01)
+
```
smanist, line 63: We don't need to do any optimization yet - it will be our next task.  For the test, just check the forward and the Jacobians.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,219 @@
...
+        self.t.add_(1)
+
+    def forward(self, state, input):
+        r'''
+        Parameters
+        ----------
+        x : Tensor
+            The state of the dynamic system
+        u : Tensor
+            The input to the dynamic system
```
smanist, line 31: Should be `state` and `input`, instead of `x` and `u`

pypose/module/dynamics.py
```diff
@@ -0,0 +1,229 @@
...
+            return self._D
+        else:
+            func = lambda x: self.observation(self.state, x)
+            return jacobian(func, self.input, **self.jacargs)
+    
+    @property
+    def c1(self):
+        if hasattr(self,'_c1'):
+            return self._c1
+
```
smanist, line 93: There should be an else as well, which returns the bias term associated with the state transition function.  Same for c2.

test/dynamics_cartpole_test.py
```diff
@@ -0,0 +1,98 @@
...
+    cartPoleSolver.set_linearization_point(jacob_state,jacob_input)
+    A = (cartPoleSolver.A).numpy()
+    B = (cartPoleSolver.B).numpy()
+    C = (cartPoleSolver.C).numpy()
+    D = (cartPoleSolver.D).numpy()
+
+    Atest = np.loadtxt("cartpoleA.txt")
+    Btest = np.loadtxt("cartpoleB.txt")
+
+    print("Error in A: ")
```
smanist, line 93: Let's try using `np.allclose` to do the comparison.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,229 @@
...
+                      The observation of the system at the current step
+        '''
+
+        new_state = self.state_transition(state, input)
+        observation = self.observation(state, input)
+        return new_state, observation
+
+    def state_transition(self):
+        pass
+
```
smanist, line 47: Better do
def state_transition(self, state, input):
    raise NotImplementedError("The users need to define their own state transition method")
Same for observation.

Also, add the documentation too.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,229 @@
...
+    def observation(self):
+        pass
+
+    def reset(self,t=0):
+        self.t.fill_(0)
```
smanist, line 52: Doc required.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,229 @@
...
+        pass
+
+    def observation(self):
+        pass
+
+    def reset(self,t=0):
+        self.t.fill_(0) 
+
+    def set_linearization_point(self, state, input):
+        self.state, self.input = state, input
```
smanist, line 55: Doc required.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @property
+    def A(self):
+        if hasattr(self, '_A'):
```
smanist, line 59: Doc required.  Same for the other properties.

test/dynamics_cartpole_test.py
smanist: Also, how about C, D, c1, and c2?

test/dynamics_lorenz_test
```diff
@@ -0,0 +1,76 @@
+import pypose.module.dynamics as ppmd
```
smanist, line 1: Perhaps we can remove this example.  In terms of nonlinear dynamics examples, cart-pole seems sufficient.

pypose/module/dynamics.py
apandey02: I'm not sure how to implement this. In the LTI system the user gives the input for c1 and c2.

test/dynamics_cartpole_test.py
apandey02: In all our nonlinear cases C is an identity matrix and D is a zero vector so I just verified that in the debugger. Can add the code for checking if required.

pypose/module/dynamics.py
apandey02: The LTI system does not define a state_transition or observation function. It will throw an error when that class is used. Should I still implement this?

test/dynamics_lorenz_test
apandey02: I thought that the idea was to have an example that doesn't require learning, and an example that requires learning. If not I can delete the example.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,368 @@
...
+    Args:
+        time (:obj:`boolean`): Whether the system is time-varying; defaults to False, meaning time-invariant
+
+    Linearization
+    ----------
+    The nonlinear state-space equation is given as:
+
+    .. math::
+        \begin{align}
+        \overrightarrow{x}_{k+1} &= \mathbf{f}(\overrightarrow{x}_k,\overrightarrow{u}_k) \\
```
smanist, line 19: Use \mathbf{} for vectors.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,368 @@
...
+        \overrightarrow{y}_{k} &= \mathbf{g}(\overrightarrow{x}_k,\overrightarrow{u}_k)
+        \end{align}
+
+    This class provides a means to linearize the system at any point along a trajectory.
+    \bf{Note}: The linearization can be provided for any arbitrary point, not just the equilibrium point(s).
+
```
smanist, line 25: See https://github.com/pypose/pypose/blob/dmpc_linearDynamics/pypose/module/dynamics.py for the syntax for note.
```
Note:
    blah
```

pypose/module/dynamics.py
```diff
@@ -0,0 +1,368 @@
...
+    Suppose we to linearize about :math:`(\overrightarrow{x}^*,\overrightarrow{u}^*)`
+    for an arbitrary equation :math:`h(x,u)`.
```
smanist, line 27: Boldface here.

smanist: Same below, for consistency.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,368 @@
...
+    Through a Taylor series expansion (ignoring higher order terms), we get the following:
+
+    .. math::
```
smanist, line 30: This is somewhat unclear.  I suggest go with the standard perturbation form:
.. math::
    h(x^*+\delta x,u^*+\delta u) = h(x^*,u^*) + \frac{\partial h}{\partial x} \right|_{x^*,u^*} \delta x + blah
Then explain that we would like to work with the full states, so we end up with
.. math::
    h(x^*+\delta x,u^*+\delta u) = \frac{\partial h}{\partial x} \right|_{x^*,u^*} (x^*+\delta x) + blah + c
where then we show the definition of c.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,368 @@
...
+                     \left. \frac{\partial h}{\partial u} \right|_{x^*,u^*} u^* + c
+    
+    Where :math:`c` is the bias generated due to the system not being at an equilibrium point.
+    If :math:`(x^*,u^*)` specify an equilibrium point, then :math:`c=0`.
+    Applying this to our state-space equations, i.e., Eqs. (1) and (2), we get:
+
+    .. math::
+        \begin{align}
+        \overrightarrow{x}_{k+1} &= \mathbf{f}(\overrightarrow{x}_k,\overrightarrow{u}_k) = 
+        \left. \frac{\partial \mathbf{f}}{\partial \overrightarrow{x}} \right|_{\overrightarrow{x}_k,\overrightarrow{u}_k} \overrightarrow{x}_k +
```
smanist, line 41: Use A, B, C, D here...

examples/module/dynamics/neuralnet.py
```diff
@@ -0,0 +1,57 @@
...
+    # Calculate trajectory
+    state_all = torch.zeros(N+1, 2)
+    state_all[0,:] = state
+    for i in range(N):
+        state_all[i+1],_ = nnSolver.forward(state_all[i], input[i])
+
+    # Create plots
+    x,y = (state_all.T).detach().numpy()
+    x_fig = createTimePlot(time,x,figname="x Plot",xlabel="Time",ylabel="x",title="x Plot")
+    y_fig = createTimePlot(time,y,figname="y Plot",xlabel="Time",ylabel="y",title="y Plot")
```
wang-chen, line 53: add space after comma `,` and between operators such as `+`.

examples/module/dynamics/neuralnet.py
```diff
@@ -0,0 +1,57 @@
...
+
+if __name__ == "__main__":
+    dt = 0.01  # Time step size
+    N  = 1000  # Number of time steps
+
+    # Time and input
+    time  = torch.arange(0, N+1) * dt
+    input = torch.sin(time)
+    # Initial state
+    state = torch.tensor((0,0))
```
wang-chen, line 39: `state = torch.tensor(0,0)` not `state = torch.tensor((0,0))`

wang-chen, line 39: do you mean `state = torch.tensor([0,0])`

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+        .. math::
+            \begin{aligned}
+            \mathbf{f}(\mathbf{x},\mathbf{u},t^*) &\approx \mathbf{f}(\mathbf{x}^*, 
+                \mathbf{u}^*,t^*) +  \left. \frac{\partial \mathbf{f}}{\partial \mathbf{x}}
+                \right|_{\chi^*} \delta \mathbf{x} + \left. \frac{\partial \mathbf{f}} 
+                {\partial \mathbf{u}} \right|_{\chi^*} \delta \mathbf{u} \\
+            &= \mathbf{f}(\mathbf{x}^*,\mathbf{u}^*,t^*) + \mathbf{A}(\mathbf{x} 
+                - \mathbf{x}^*) + \mathbf{B}(\mathbf{u}-\mathbf{u}^*) \\
+            &= \mathbf{A}\mathbf{x} + \mathbf{B}\mathbf{u} + \mathbf{c}_1
+            \end{aligned}
```
wang-chen, line 54: I moved the mathematical definition of `A`, `B` ... to their properties

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+        input : Tensor
+                The input to the dynamical system
+        t     : Tensor
+                The time step of the dynamical system
+
+        Returns
+        ----------
+        None
+        '''
+        self.state, self.input, self.t = state, input, t
```
wang-chen, line 154: This [forward_hook line](https://github.com/pypose/pypose/blob/82602d76014c9291533d9811c2aea3f6143f3fcc/pypose/module/dynamics.py#L83) was originally for this purpose. Maybe we should add an argument to `forward` such as `time=None`, so that time can also be stored.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+        if self.A.ndim >= 3:
+            assert self.A.ndim == state.ndim == input.ndim,  "Invalid System Matrices dimensions"
+        else:
+            assert self.A.ndim == 2,  "Invalid System Matrices dimensions"
+
+        z = self.state_transition(state, input)
+        y = self.observation(state, input)
+        return z, y
+
+    def state_transition(self, state, input):
```
wang-chen, line 338: add a short doc:
```
Perform one step for linear state transition.
```
maybe we can also move the linear state transition equation here.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+        return state.matmul(self.A.mT) + input.matmul(self.B.mT) + self.c1
+
+    def observation(self, state, input):
```
wang-chen, line 341: add a short doc:
```
Perform one step for linear system observation.
```
maybe we can also move the linear observation function here.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
+import torch as torch
+import torch.nn as nn
+import pypose as pp
+from torch.autograd.functional import jacobian
+
+
+class System(nn.Module):
+    r'''
+    The general base class of discrete system dynamics model.
```
wang-chen, line 9: The general base class of discrete-time system dynamics model.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Example:
+            >>> A = torch.randn((2,3,3))
+                B = torch.randn((2,3,2))
+                C = torch.randn((2,3,3))
+                D = torch.randn((2,3,2))
+                c1 = torch.randn((2,1,3))
+                c2 = torch.randn((2,1,3))
+                state = torch.randn((2,1,3))
+                input = torch.randn((2,1,2))
```
wang-chen, line 318: you need `>>>` for all those lines.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+            tensor([[[-3.5780, -2.2970, -2.9314]],
+                    [[-0.4358,  1.7306,  2.7514]]]))
+
+        Note:
+            In this general example, all variables are in a batch. User definable as appropriate.
+        '''
+        if self.A.ndim >= 3:
+            assert self.A.ndim == state.ndim == input.ndim,  "Invalid System Matrices dimensions"
+        else:
+            assert self.A.ndim == 2,  "Invalid System Matrices dimensions"
```
wang-chen, line 332: this should be moved to the `__init__`. As a result, this forward function doesn't need to be defined here, as its totally same with the parent class.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+    def c2(self, c2):
+        self._c2 = c2
+
+
+class LTI(System):
+    r'''
+    Discrete-time Linear Time-Invariant (LTI) system.
+    
+    Args:
+        A, B, C, D (:obj:`Tensor`): The coefficient matrix in the state-space equation of LTI system,
```
wang-chen, line 268: define seperately.
A, B, C, D have their own name, e.g., state transition matrix.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,342 @@
...
+        c1 (:obj:`Tensor`): The constant input of the system,
+        c2 (:obj:`Tensor`): The constant output of the system,
+        state (:obj:`Tensor`): The state of the current timestamp of LTI system,
+        input (:obj:`Tensor`): The input of the current timestamp of LTI system.
```
wang-chen, line 272: you don't have `state` and `input` arguments in `__init__`

test/module/test_dynamics.py
```diff
@@ -0,0 +1,192 @@
...
+    N: channels
+
+    A = torch.randn((N,n,n))
+    B = torch.randn((N,n,p))
+    C = torch.randn((N,q,n))
+    D = torch.randn((N,q,p))
+    c1 = torch.randn(N,1,n)
+    c2 = torch.randn(N,1,q)
+    state = torch.randn((N,1,n))
+    input = torch.randn((N,1,p))
```
wang-chen, line 116:
```
A = torch.randn(N, n, n)
B = torch.randn(N, n, p)
```
not
```
A = torch.randn((N,n,n))
B = torch.randn((N,n,p))
```
BTW, remember space after all commas.

examples/module/dynamics/neuralnet.py
```diff
@@ -0,0 +1,57 @@
...
+        super().__init__(time)
+        self.net = torch.nn.Sequential(
+            torch.nn.Linear(2,hiddenSize[0]),
+            torch.nn.ReLU(),
+            torch.nn.Linear(hiddenSize[0],hiddenSize[1]),
+            torch.nn.ReLU(),
+            torch.nn.Linear(hiddenSize[1],2))
+    
+    def state_transition(self, state, input):
+        return self.net(state)+input
```
wang-chen, line 18: remember spaces before and after all operators.

wang-chen (changes requested):
All lines should not be longer than 90-100 characters.

pypose/module/dynamics.py
wang-chen: I am cosidering removing `forward_hook` or this `set_linearization_point`, as they are repeated.

wang-chen:
We need an example in doc of `System` for defining a simple time-variant and time-invariant non-linear system.

pypose/module/dynamics.py
smanist: `forward_hook` indeed could be removed, as it might cause changes in the data members that are not visible to the users.  For `set_linearization_point`, my concern is whether one would like to involve `forward` where the back-propagation is not needed.  How about making a function called `set_states`, that does what's in 154, and then the `forward` can call this function before the time stepping, while if one just wants to do linearization, s/he can just call `set_states`.

pypose/module/dynamics.py
smanist: Agreed.

pypose/module/dynamics.py
wang-chen: How about we change to something like this:

```
class System(nn.Module):
    def __init__(self):
        super().__init__()
        self.jacargs = {'vectorize':True, 'strategy':'reverse-mode'}

    def forward(self, state, input, time=None):
        self.state, self.input = input, state
        state = self.state_transition(state, input, time)
        output = self.observation(state, input, time)
        if time is not None:
            return state, output, time + 1
        else:
            return state, output

    def state_transition(self, state, input, time=None):
        raise NotImplementedError("The users need to define their own state transition method")

    def observation(self, state, input, time=None):
        raise NotImplementedError("The users need to define their own observation method")

    def reset(self, t=0):
        self.t.fill_(t)

    @property
    def A(self):
        return self.A_(self.state, self.input)

    @A.setter
    def A(self, A):
        self._A = A

    def A_(self, state, input):
        if hasattr(self, '_A'):
            return self._A
        func = lambda x: self.state_transition(x, input)
        return jacobian(func, state, **self.jacargs)
```
In this case, the user can use `A` directly with `mysystem.A`, which in default linearize the last system point.
Or if a linearization point is needed then user can call `mysystem.A_(state=state, input=input)`. 
Then we don't need to set states or inputs.

smanist: I think it is a good idea to have `A_` as you defined, though `time` should be in the input as well.
As for the `forward`, perhaps we can replace `reset` with `step_counter` to encapsulate the changes in time?
```
    def forward(self, state, input, time=None):
        self.state, self.input = input, state
        self.step_counter(time)
        new_state = self.state_transition(self.state, self.input, self.time)
        output = self.observation(self.state, self.input, self.time)
        return new_state, output, self.time
    def step_counter(self, time=None):
        self.time = self.time+1 if time is None else time
        return self.time
```

wang-chen: 1. when `time` is not passed to `forward`, it shouldn't return `time`. But in this example, it always returns `time`, so that the user has to always use `x, y, _ = system(x, u)` instead of `x, y = system(x, u)`.
2. the `state_transition` function should use current `time`, not `next time`? but `self.step_counter` is called before `state_transition`.
3. if it is called `step_counter`, then we should use `step` instead of `time`?
4. The original `reset` is to let the user restart a system in their program, not for the counter.

smanist: After the Zoom discussion, I tried to implement the changes.  But what we discussed still might not make sense...

Consider this example:
```
class Floquet(System):
    def __init__(self):
        super(Floquet, self).__init__(time=True)

    def state_transition(self, state, input, t=None):
        cc = torch.cos(2*math.pi*self._t/10)
        ss = torch.sin(2*math.pi*self._t/10)
        A = torch.tensor([
            [1., cc/10],
            [cc/10, 1.]])
        B = torch.tensor([
            [ss],
            [1.]])
        return (A.matmul(state.reshape(-1,1)) + B * input).squeeze()
```

By default we let the user use `self._t` as the time variable in `state_transition`/`observation`.  This actually effectively makes the argument `t` useless.  When linearizing, the code would end up doing
```
        func = lambda x: self.state_transition(x, self.input, self.time)
        return jacobian(func, self.state, **self.jacargs)
```
but the `self.time` would not be used at all, which defeats the purpose of using `set_ref_point`.

I might be missing something...

smanist: One possibility is to -
```
    def forward(self, state, input):
        self.state, self.input = state, input
        return self.state_transition(state, input, self._t), self.observation(state, input, self._t)
    def state_transition(self, state, input, t):
        blah
```
so that `forward` always uses `self._t`, which is the system time.  But the user has to use the argument `t` of `state_transition` as the time variable, instead of `self._t`.

BTW, it seems now `time=False` in `__init__` is meaningless - the user just defines `state_transition` and whether the system is time-varying just depends on whether they use `t`.

smanist: Implemented and tested what I suggested - see my latest commit.

wang-chen (approved):

PR #100, opened 2022-07-26
Dmpc linear dynamics

hakunanh:
updated LTI system


pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        if hasattr(self, '_D'):
+            return self._D
+        else:
+            func = lambda x: self.observation(self.state, x)
+            return jacobian(func, self.input, **self.jacargs)
+
+
+class LTI(_System):
+    r'''
+    A sub-class of '_System' to represent Linear Time-Invariant system.
```
aerogjy, line 67: represent the dynamics of Linear Time-Invariant (LTI) system.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+    
+    Args:
+        A, B, C, D (:obj:`Tensor`): The coefficient matrix in the state-space equation of LTI system
+        c1, c2 (:obj:`Tensor`): Bias generated by system
```
aerogjy, line 71: Let's not say bias. Instead, constant input and output of the system.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        
+    Note:
+        According to the actual physical meaning, the dimensions of A, B, C, D must be the same,
```
aerogjy, line 74: same -> consistent

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        whether in the batch case or not.
```
aerogjy, line 75: whether in batch or not.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        return self._c1
+    
+    @property
+    def c2(self):
+        return self._c2
+    
+    def forward(self, x, u):
+        r'''
+        Args:
+            x (:obj:`Tensor`): The state of LTI system,
```
aerogjy, line 96: The state of the current timestamp of LTI system.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+            u (:obj:`Tensor`): The input of LTI system.
```
aerogjy, line 97: The input of the current timestamp of LTI system.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Return:
+            Tensor: The solved batched tensor, which means state-transition and observation of the system.
```
aerogjy, line 100: The state of the next timestamp (state-transition) and the system output (observation). 
Return type: should be "tuple of tensors". (you have both the state and observation)

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+
+            
+        Every linear time-invariant lumped system can be described by a set of equations of the form
```
aerogjy, line 103: Let the current state of the system be :math:`\mathbf{x}`, input be :math:`\mathbf{u}'; the state of next time stamp be :math:`\mathbf{z}`, system output be :math:`\mathbf{y}`; system constant input and output be :math:`\mathbf{c}_1` and :math:`\mathbf{c}_2`. The LTI system can be described by

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        which is called the state-space equation.
+        
+        .. math::
+            \begin{align*}
+                z_{i} = A_{i} \times x_{i} + B_{i} \times u_{i} + c_1 \\
```
aerogjy, line 108: Let's make the vector and matrix in bold. the `\times` can be ignored.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+                y_{i} = C_{i} \times x_{i} + D_{i} \times u_{i} + c_2 \\
+            \end{align*}
+            
+        where :math:`x` is actually :math:`\dot{x}`, the differential form of :math:`x`.
```
aerogjy, line 112: Replace this line by "Here, we consider the discrete system dynamics. "

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        
+        Let the input be matrix :math:`A`, :math:`B`, :math:`C`, :math:`D`, :math:`x`, :math:`u`, :math:`c_1`, :math:`c_2`.
```
aerogjy, line 114: Line 114, 115 can be deleted.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+                z_{i} = A_{i} \times x_{i} + B_{i} \times u_{i} + c_1 \\
+                y_{i} = C_{i} \times x_{i} + D_{i} \times u_{i} + c_2 \\
+            \end{align*}
+            
+        where :math:`x` is actually :math:`\dot{x}`, the differential form of :math:`x`.
+        
+        Let the input be matrix :math:`A`, :math:`B`, :math:`C`, :math:`D`, :math:`x`, :math:`u`, :math:`c_1`, :math:`c_2`. 
+        :math:`x_i` represents each individual matrix in the batch. 
+        
+        Note:
```
aerogjy, line 117: I'm glad to see you put these details in this note. That means you are clear about them. Great! 
However, I think for this software doc, we don't have to provide these kind of details. Common users should be able to figure them out. 
I suggest we only keep the item " x, u, A, B, C, D can be single input or in a batch. In the batch case, their dimension must be consistent." You can polish this sentence. Make sure the symbols are in bold.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+        Example:
+            >>> A = torch.randn((2,3,3))
+                B = torch.randn((2,3,2))
+                C = torch.randn((2,3,3))
+                D = torch.randn((2,3,2))
+                c1 = torch.randn((2,1,3))
+                c2 = torch.randn((2,1,3))
+                x = torch.randn((2,1,3))
+                u = torch.randn((2,1,2))
+            >>> A
```
aerogjy, line 137: In this example, it is not necessary to print out A, B, C, D, c1, c2, x, u explicitly. User can do them if they want. Let's just keep the randn part and the final lti part.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+                c2
+            tensor([[[-0.6067, -1.8852, -0.8722]],
+                    [[-0.3974,  2.1444,  0.5077]]]) 
+                x
+            tensor([[[ 0.9576,  0.6824, -2.3676]],
+                    [[-0.5359,  1.4186,  0.9334]]])  
+                u
+            tensor([[[ 1.0024,  1.9124]],
+                    [[-0.9396,  1.1412]]])
+            >>> lti = LTI(A, B, C, D, c1, c2)
```
aerogjy, line 177: This example is inconsistent with that in the test script. 
lti = pp.module.LTI(...)
lti(x,u)

test/dynamics_test.py
```diff
@@ -0,0 +1,39 @@
+import sys
+sys.path.append("..")
+import torch as torch
+import pypose as pp
+import torch.nn as nn
+
+"""
+A(n*n), B(n*p), C(q*n), D(q*p)
+x(n*1), u(p*1), N: channels
+A = torch.randn((n,n))
```
aerogjy, line 10: A, B, C, D should also be (N, n, n), etc?

test/dynamics_test.py
```diff
@@ -0,0 +1,39 @@
...
+x = torch.randn((5,1,4))
+u = torch.randn((5,1,2))
+
+lti = pp.module.LTI(A, B, C, D, c1, c2)
+
+#print(A,B,C,D,c1,c2,x,u)    
+#The user can implement this line to print each parameter for comparison.
+
+print(lti(x,u))
+
```
aerogjy, line 35: Should we also give a non-batch example?

pypose/module/dynamics.py
```diff
@@ -0,0 +1,196 @@
...
+    
+        Note:
+            In this general example, all variables are in the batch. User definable as appropriate.
+            
+        '''
+
+        if self.A.ndim >= 3:
+            assert self.A.ndim == x.ndim == u.ndim,  "Invalid System Matrices dimensions"
+        else:
+            assert self.A.ndim == 2,  "Invalid System Matrices dimensions"
```
aerogjy, line 191: Do you mean A has to be in batch?

aerogjy (commented):
Doc reviewed.

aerogjy (approved):
I directly polished some writing.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,158 @@
...
+        which is called the state-space equation.
+        
+        .. math::
+            \begin{align*}
+                \mathbf{z} = \mathbf{A}\mathbf{x} + \mathbf{B}\mathbf{u} + \mathbf{c}_1 \\
+                \mathbf{y} = \mathbf{C}\mathbf{x} + \mathbf{D}\mathbf{u} + \mathbf{c}_2 \\
+            \end{align*}
+
+        where we use :math:`\mathbf{x}` and :math:`\mathbf{u}` to represent state and input of the current timestamp of LTI system.
+
```
wang-chen, line 113: 1. Will it be better if moving this equation and explanation to the constructor of this class? Then the forward function only needs to provide examples. 
2. The arguments of the constructor will be self-explained also.

pypose/module/dynamics.py
```diff
@@ -0,0 +1,158 @@
...
+        y = self.observation(state, input)
+        return z, y
+
+    def state_trasition(self, state, input):
+        state_transition = state.matmul(self._A.mT) + input.matmul(self._B.mT) + self._c1
+        return state_transition
+
+    def observation(self, state, input):
+        observation = state.matmul(self._C.mT) + input.matmul(self._D.mT) + self._c2
+        return observation
```
wang-chen, line 158: Suggested to change to
```
return state.matmul(self._C.mT) + input.matmul(self._D.mT) + self._c2
```

wang-chen:
I just found another big problem, remember that you have set properties `A`, `B`, `C`, and `D` in base class, so you shouldn't use `self._A` in the subclass, you should only use 'self.A' in a subclass. Moreover, becasue you also need to set the value for those properties, so you need to define property setter. You may follow [this](https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters).

wang-chen:
> > I just found another big problem, remember that you have set properties `A`, `B`, `C`, and `D` in base class, so you shouldn't use `self._A` in the subclass, you should only use 'self.A' in a subclass. Moreover, becasue you also need to set the value for those properties, so you need to define property setter. You may follow [this](https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters).
> 
> Thanks for your feedback. I have a question with the branch. Since you have merged the main branch into this pr, what should I do after I have modified the original and the test file? Should I still push to my own branch and re-request review？

Yes, you still push to this branch and we will merge this pr into the main branch finally after you make all changes and re-request a review.

PR #130, opened 2022-09-21
expand first column in docs

ew123:


bradygm:
Seems to have fixed itself somehow...

PR #134, opened 2022-09-22
kernel function

qinhanhahaha:


wang-chen (changes requested):
1. Please remove the unnecessary folder and files: `.idea/inspectionProfiles`
2. For all figures, could you make sure the grid of x, y axes are equal, e.g., `axis('equal')`
3. Could you plot at least 2 curves for each figure for different hyperparameters, e.g., for `Scale`, draw two curves for  `delta = 0.8` and `delta = 0.6`? Remember to add a legend for each curve.

pypose/optim/kernel.py
```diff
@@ -123,3 +132,168 @@ def forward(self, input: Tensor) -> Tensor:
...
+        input.
+
+    Example:
+        >>> import pypose.optim.kernel as ppok
+        >>> kernel = ppok.Arctan()
+        >>> input = torch.tensor([0, 0.5, 1, 2, 3])
+        >>> kernel(input)
+        tensor([0.0000, 0.4636, 0.7854, 1.1071, 1.2490])
+
+    .. figure:: /_static/img/module/Kernal/Arctan.png
```
wang-chen, line 201: Better to use parameter `a, b = 1.0, -1.0`, `a, b = 2.0, -1.0`, and `a, b = 2.0, -2.0`, so that the users can find the effect of `a` and `b`, respectively. For example, 

- the effect of `a` can be found by comparing curve of `a, b = 1.0, -1.0` and  `a, b = 2.0, -1.0`.
-  the effect of `b` can be found by comparing curve of `a, b = 2.0, -1.0`, and `a, b = 2.0, -2.0`.
- Remember add space after the comma `,`, add space before and after operator `=`.
- use `1.0` instead of `1` to have the same width for all legends.

wang-chen (changes requested):
- Hi, you removed the unnecessary files added last time, but this time you added more unnecessary files in folder `.idea`.  Please remove them.
- please change `kernal` to `kernel` in all figure titles, folder names, and linked paths, etc.
- move figures to 'docs/source/_static/img/optim/kernel/'.

pypose/optim/kernel.py
wang-chen: you'd better change all figure file names to small case, e.g., `scale.png` instead of `Scale.png`.

qinhanhahaha:
please have a check

wang-chen (changes requested):
One final change request:
The space should be after the comma, not before the comma in the second legend of the Tolerant Kernel, i.e., `a, b = 2.0, -1.0` instead of `a, b = 2.0 ,-1.0`
image

wang-chen (approved):

PR #139, opened 2022-10-05
Imu int

haleqiu:
Reduce one redundant step && Modify the inner function of the imu preintegrator to easy the process of calling the function outside


wang-chen:
Do you  @haleqiu have still need to merge this PR? Is it ready for review?

wang-chen (approved):

PR #144, opened 2022-10-13
icp

qinhanhahaha:
icp_module


wang-chen (changes requested):
Some quick comments you have to address.

- You cannot use libraries like `numpy` and `sklearn`, they are not differentiable. You have to implement all the operations using `torch` and `pypose`.
- Please use `LieTensor` such as `SO3` and `SE3` to represent transform and rotations.
- Remember to add documentation and tests.

wang-chen:
closed as no more updates

PR #155, opened 2022-10-28
weight matrix in optimization

xukuanHIT:


pypose/optim/optimizer.py
```diff
@@ -371,7 +378,8 @@ def __init__(self, model, solver=None, strategy=None, kernel=None, corrector=Non
         self.strategy = TrustRegion() if strategy is None else strategy
         defaults = {**{'min':min, 'max':max}, **self.strategy.defaults}
         super().__init__(model.parameters(), defaults=defaults)
-        self.jackwargs = {'vectorize': vectorize, 'flatten': True}
+        # self.jackwargs = {'vectorize': vectorize, 'flatten': True}
+        self.jackwargs = {'vectorize': vectorize, 'flatten': False}
```
wang-chen, line 382: `flatten` is `False` doesn't support multiple Model parameters, please change back to `True`. will add an additional test for this.

pypose/optim/optimizer.py
```diff
@@ -56,14 +56,15 @@ def model_forward(self, input):
...
-        else:
-            residual = error
-        return residual.squeeze(-1)
+        # error = (output if target is None else output - target).unsqueeze(-1)
+        # if weight is not None:
+        #     residual = self.validate(weight) @ error
+        # elif hasattr(self, 'weight'):
+        #     residual = self.weight @ error
+        # else:
+        #     residual = error
```
wang-chen, line 65: remove unnecessary lines.

wang-chen (changes requested):
1. remove all unnecessary lines including commented lines and `RobustModel.validate`
2. `weight` can also be given to `optimizer`, and now `Robust Model` doesn't accept `weight` anymore, you need to register 'weight' it in `optimizer`.
3. `[i for i in range(len(J.shape)-1)]` is `list(range(N-1))`
4. `J.reshape` will copy memory, do you really need that?
5. Update documentation as modified.

pypose/optim/optimizer.py
```diff
@@ -451,17 +434,23 @@ def step(self, input, target=None, weight=None):
...
         for pg in self.param_groups:
+            weight = self.weight if weight is None else weight
             R = self.model(input, target, weight)
             J = modjac(self.model, input=(input, target, weight), **self.jackwargs)
             R, J = self.corrector(R = R, J = J)
             self.last = self.loss = self.loss if hasattr(self, 'loss') \
                                     else self.model.loss(input, target, weight)
-            A, self.reject_count = J.T @ J, 0
+            J_T = J.reshape(tuple(R.shape)+(-1,))
+            J_T = J_T.permute([len(J_T.shape)-1,]+list(range(len(J_T.shape)-1)))
```
wang-chen, line 444: why do you want the dimension of J to be rotated?
and `len(J_T.shape)` is `J_T.ndim`

wang-chen (changes requested):

pypose/optim/optimizer.py
xukuanHIT: > len(

Screenshot from 2022-12-17 16-59-34

pypose/optim/optimizer.py
```diff
@@ -451,17 +434,28 @@ def step(self, input, target=None, weight=None):
...
         for pg in self.param_groups:
+            weight = self.weight if weight is None else weight
             R = self.model(input, target, weight)
             J = modjac(self.model, input=(input, target, weight), **self.jackwargs)
             R, J = self.corrector(R = R, J = J)
             self.last = self.loss = self.loss if hasattr(self, 'loss') \
                                     else self.model.loss(input, target, weight)
-            A, self.reject_count = J.T @ J, 0
+            # Let Nr: number of residuals, Dr: dimension of each residual, Np: number of model parameters
+            # Now, the shape of R: [Nr, Dr], J: [Nr*Dr, Np], weight: [Nr, Dr, Dr]
```
wang-chen, line 444: This is not correct, `R` can be any higher dimension, depending on the model defined by the users. I will and a test example to the test folder.

xukuanHIT:
The reason why the loss didn't decrease in the previous tests is that we can get the optimal solution in one step. Now I have changed "test_optim_anybatch" to "multi-parameter", so we can see the explicit loss decrease.

pypose/optim/optimizer.py
```diff
@@ -451,22 +434,28 @@ def step(self, input, target=None, weight=None):
...
+            weight = self.weight if weight is None else weight
+            R = self.model(input, target)
+            J = modjac(self.model, input=(input, target), **self.jackwargs)
             R, J = self.corrector(R = R, J = J)
             self.last = self.loss = self.loss if hasattr(self, 'loss') \
-                                    else self.model.loss(input, target, weight)
-            A, self.reject_count = J.T @ J, 0
+                                    else self.model.loss(input, target)
+            J_T = J.reshape(tuple(R.shape)+(-1,))
+            J_T = J_T.permute([J_T.ndim-1,]+list(range(J_T.ndim-1)))
```
wang-chen, line 444: How about changing those two lines to:
```python
J_T = J.T.reshape((-1,) + R.shape)
```

wang-chen (approved):

PR #163, opened 2022-12-13
add euler function

wang-chen:
add `euler` for any `LieTensor` types.


haleqiu (changes requested):
Two problems were observed:
1. the formulas `t1` and `t4` in `euler` assume that the inputted quaternion is normalized, which is not true.
2. the singularity of Euler is not solved.

haleqiu (approved):

PR #169, opened 2023-01-05
Differential MPC

hakunanh:
lqr and train mpc with running pytest successfully


wang-chen:
@hakunanh please look at my [changes](https://github.com/pypose/pypose/pull/169/commits/2523e384bd1ebcae2d56441f12ce31629715c293), please
1. Verify changes in `pypose.module.LQR` and check the results with your own tests.
2. Update doc for LQR with the list of a recursive algorithm.
3. Update test for MPC.

pypose/module/dynamics.py
```diff
@@ -6,7 +6,587 @@
...
+
+        Warning:
+            For nonlinear systems, the users have to call this function before getting the
+            linearized system.
+        '''
+        self.systime = t
+        return self
+
+
+class NTI(System):
```
aerogjy, line 389: @wang-chen I completely don't like the idea of separating the nonlinear into NTI, NTV, people never call these concepts at all in general. Even in your code implementation, there are also not much difference between those two with the only difference is the systime t, which shouldn't be ignored even under your concept of NTI, because for a nonlinear system, time can never be ignored, unlike the LTI you can have some common properties, in nonlinear system, time really matter. User usually want to reset time, re-initialize the x(0), or even split a slice of state at some specific t. 
Also, if this is the case, why are we still leave the set_refpoint in the base class. In short , I still think the previous based class with set_refpoint and systime makes more sense. And as a system, A,B,C,D either from direct user input or jacobian linearization are important properties there.

pypose/module/dynamics.py
```diff
@@ -6,7 +6,587 @@
...
+        return self._c1
+
+    @property
+    def c2(self):
+        r'''Constant term generated by observation.
+        '''
+        return self._c2
+
+
+class LTV(LTI):
```
aerogjy, line 230: @wang-chen I'm not sure whether I like this idea. My first glance is oh, we provide a LTV system for the user. Then, after looking into this class, I found not much difference with our previous base class of system. People sill need to inherit from this class to create their own. So I kind of think there is no necessary to do this at all. (This just add an addition layer in the middle without actually do much.). Also, the A, B, C, D properties interface seem to be missing, which should be provided.

pypose/module/dynamics.py
```diff
@@ -6,7 +6,587 @@
...
+        ...     def __init__(self):
+        ...         super().__init__()
+        ... 
+        ...     def state_transition(self, state, input):
+        ...         return state.cos() + input.sin()
+        ... 
+        ...     def observation(self, state, input):
+        ...         return state
+        ... 
+        >>> lti = MyNTI()
```
aerogjy, line 466: nti

pypose/module/dynamics.py
```diff
@@ -6,7 +6,587 @@
...
+        >>> current, input = torch.randn(3), torch.randn(3)
+        >>> next, observation = lti(current, input)
```
aerogjy, line 468: nti

aerogjy (changes requested):
I spent time on lqr, test_lqr, and dynamics. I haven't looked at mpc related stuff in the examples. In general, some minor issues in the lqr with the biggest concern is the class input, which I think, system, Q, p, T should all be the input when defining the lqr class rather than only the system. Others things are just documentation. I really like the current concise version, I also used to define variable first, then allocation number rather than dynamically allocate tensors. Thanks @wang-chen for providing a professional example, I think @hakunanh will learn from this from now on.
As for the dynamics, @wang-chen I apologize for the slow review. But after the deep look, I really couldn't agree the current separation of NTI, NTV, these conceps are so strange, and it's not necessary, systime t matter a lot for any nonlinear system. (Please see my detailed comments)

examples/module/mpc/train_mpc.py
```diff
@@ -0,0 +1,125 @@
...
+    loss_f.write('im_loss,mse\n')
+    loss_f.flush()
+    fname = os.path.join(args.save, 'pypose time.csv')
+    time_d = open(fname, 'w')
+    time_d.write('backward_time,overall_time\n')
+    time_d.flush()
+
+    def get_loss(x_init, _A, _B):
+        lti = pp.module.LTI(expert['A'], expert['B'], C, D, c1, c2)
+        _lqr2 = mpc.MPC(lti, u_init=u_init, lqr_iter=1)
```
aerogjy, line 77: why naming it as _lqr2 twice? why not mpc: mpc_expert and mpc_agent?

examples/module/mpc/train_mpc.py
```diff
@@ -0,0 +1,125 @@
...
+                     torch.mean((B - expert['B'])**2)
+
+        loss_f.write('{},{}\n'.format(traj_loss.item(), model_loss.item()))
+        loss_f.flush()
+        time_d.write('{},{}\n'.format(backward_time, overall_time))
+        time_d.flush()
+
+        plot_interval = 100
+        if i % plot_interval == 0:
+            os.system('./plot.py "{}" &'.format(args.save))
```
aerogjy, line 115: no plot.py script.

examples/module/mpc/mpc.py
```diff
@@ -0,0 +1,74 @@
...
+
+        n_not_improved = 0
+        for i in range(self.lqr_iter):
+            if u is not None or not u.requires_grad:
+                u = u
+            else:
+                u = u.detach()
+            u.requires_grad = False 
+
+            _lqr = pp.module.LQR(system = self.system)
```
aerogjy, line 44: here should adapt to the new lqr interface, right?

examples/module/mpc/train_mpc.py
```diff
@@ -0,0 +1,125 @@
...
+    loss_f = open(fname, 'w')
+    loss_f.write('im_loss,mse\n')
+    loss_f.flush()
+    fname = os.path.join(args.save, 'pypose time.csv')
+    time_d = open(fname, 'w')
+    time_d.write('backward_time,overall_time\n')
+    time_d.flush()
+
+    def get_loss(x_init, _A, _B):
+        lti = pp.module.LTI(expert['A'], expert['B'], C, D, c1, c2)
```
aerogjy, line 76: the expert system only need to be generate once, out side the loop. Why don't we generate that right after defining the expert dict?

examples/module/mpc/train_mpc.py
```diff
@@ -0,0 +1,125 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--n_batch', type=int, default=16)
+    parser.add_argument('--n_state', type=int, default=5)
+    parser.add_argument('--n_ctrl', type=int, default=5)
+    parser.add_argument('--T', type=int, default=5)
+    parser.add_argument('--save', type=str)
+    parser.add_argument('--work', type=str, default='work')
+    parser.add_argument('--no-cuda', action='store_true')
```
aerogjy, line 20: right now, I can only get the --no-cuda option work....the gpu option always complaining about some variables are on different devices., e.g. L136 in lqr.py

wang-chen (approved):

wang-chen (approved):

wang-chen (approved):

PR #182, opened 2023-02-04
PF model

HongLouyemeng:
1.add PF model.


HongLouyemeng:
> 1.add PF model.

This version still gives random results when the input and particle are the same.

pypose/module/pf.py
```diff
@@ -0,0 +1,204 @@
...
+            R (:obj:`Tensor`, optional): covariance of system observation model
+
+        Return:
+            list of :obj:`Tensor`: posteriori state and covariance estimation
+        '''
+        # Upper cases are matrices, lower cases are vectors
+
+        Q = Q if Q is not None else self.Q
+        R = R if R is not None else self.R
+        conv_weight = len(x) if conv_weight is None else conv_weight
```
wang-chen, line 162: this is different from doc.
Do you mean `x.size(-1)` instead of `len(x)`?
Do you mean `cov_weight` instead of `conv_weight`?
Is it better to just define `conv_weight` as `n`, which will be consistent with the doc?

pypose/module/pf.py
```diff
@@ -0,0 +1,204 @@
...
+        q = q / torch.sum(q)
+        return q
+
+    def resample_particles(self, q, x):
+        r'''
+        Resample the set of a posteriori particles
+        '''
+        r = torch.rand(self.particle_number, device=x.device)
+        cumsumq = torch.cumsum(q, dim=0)
+        cumsumq[-1] = 1.0
```
wang-chen, line 198: do you want `torch.testing.assert_allclose(cumsumq[-1], torch.tensor(1.0, device=r.device))`

pypose/module/pf.py
```diff
@@ -0,0 +1,204 @@
...
+        '''
+        q = torch.distributions.MultivariateNormal(ye, R).log_prob(y).exp()
+        q = q / torch.sum(q)
+        return q
+
+    def resample_particles(self, q, x):
+        r'''
+        Resample the set of a posteriori particles
+        '''
+        r = torch.rand(self.particle_number, device=x.device)
```
wang-chen: do you need `r = torch.rand(self.particles, dtype=x.dtype, device=x.device)`

pypose/module/pf.py
```diff
@@ -0,0 +1,204 @@
...
+        * Dan Simon, `Optimal State Estimation: Kalman, H∞, and Nonlinear Approaches
+          <https://onlinelibrary.wiley.com/doi/epdf/10.1002/0470045345.fmatter>`_,
+          Cleveland State University, 2006
+    '''
+
+    def __init__(self, model, Q=None, R=None, particle_number=None):
+        super().__init__(model, Q, R)
+        self.particle_number = 1000 if particle_number is None else particle_number
+
+    def forward(self, x, y, u, P, Q=None, R=None, t=None, conv_weight=None):
```
wang-chen, line 143: what is `conv_weight`? this is not defined in the doc Args. is it `cov_weight`

wang-chen (changes requested):

pypose/module/pf.py
```diff
@@ -0,0 +1,204 @@
...
+        '''
+        r = torch.rand(self.particles, dtype=x.dtype, device=x.device)
+        cumsumq = torch.cumsum(q, dim=0)
+        cumsumq[-1] = 1.0
+        return x[torch.searchsorted(cumsumq, r)]
+
+    def compute_cov(self, a, b, w, Q=0):
+        '''Compute covariance of two set of variables.'''
+        a, b = a.unsqueeze(-1), b.unsqueeze(-1)
+        return Q + (w.unsqueeze(-1) * a @ b.mT).sum(dim=-3)
```
wang-chen, line 204: do you mean
```
from torch import cross
return Q + cross(w * a, b).sum(-3)
```

pypose/module/pf.py
HongLouyemeng: > assert_allclose
**log_prob.exp()** will get the **None** list, we  need to assign -1 to avoid the denominator to be None
```python
 q = MultivariateNormal(ye, R).log_prob(y).exp()
```

pypose/module/pf.py
HongLouyemeng: Maybe **torch.cross** doesn't work.

pypose/module/pf.py
```diff
@@ -0,0 +1,212 @@
...
+        '''
+        m = MultivariateNormal(x, P)
+        return m.sample(torch.Size([self.particles]))
+
+    def relative_likelihood(self, y, ye, R):
+        r'''
+        Compute the relative likelihood
+        '''
+        q = MultivariateNormal(ye, R).log_prob(y).exp()
+        return F.normalize(q, p=1, dim=-1)
```
wang-chen, line 198: How about
```python
return F.softmax(MultivariateNormal(ye, R).log_prob(y), dim=-1)
```
This could be more numerically stable and may not produce any `Nan`.

pypose/module/pf.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def resample_particles(self, q, x):
+        r'''
+        Resample the set of a posteriori particles
+        '''
+        r = torch.rand(self.particles, dtype=x.dtype, device=x.device)
+        cumsumq = torch.cumsum(q, dim=0)
+        cumsumq[-1] = 1.0
```
wang-chen, line 206: if the above works, then this setting to `1` will not needed.

wang-chen: It is a pity that `MultivariateNormal` subclasses [`Distribution`](https://pytorch.org/docs/stable/distributions.html?highlight=distribution#torch.distributions.distribution.Distribution), but doesn't implement `cdf`, otherwise we can combine `relative_likelihood` and `resample_particles` into one function like this:
```python
def resample_relative_likelihood(self, y, ye, R, x):
    r = torch.rand(self.particles, dtype=x.dtype, device=x.device)
    cumsumq = MultivariateNormal(ye, R).cdf(y)
    return x[torch.searchsorted(cumsumq, r)]
```

wang-chen (changes requested):

pypose/module/pf.py
HongLouyemeng: Sorry to see it so late, thanks for your advice.

pypose/module/pf.py
HongLouyemeng: It works.

wang-chen (approved):

PR #188, opened 2023-03-04
update to scheduler.continual()

wang-chen:
Sat Mar. 4th update:
We are planning a change `scheduler.continual` to `scheduler.continual()`, do you agree with this change? This update will raise an error, if the users still call `scheduler.continual`.

Expected usage:
image

Usage that will trigger a warning:
image

New error msg will be:
image

Old content, originally we were thinking of changing it to iscontinual():
Pasted Graphic
Pasted Graphic 1
image

Actually, we prefer to change to `scheduler.continual()`. However, Python doesn’t allow a property and a method have the same name, thus this error msg will not be raised if we simply change to `scheduler.continual()`. In this case, the loop will not be stopped if the users don’t know this change.

Do you have better suggestions that we can change to `scheduler.continual()` but still keep the users informed?

The reason that we want make this change is that we may run more checking and others in a future release.

Old code example is [here](https://github.com/pypose/pypose/blob/main/test/optim/test_scheduler.py)


zitongzhan:
Pushed my solution, and updated PR description as well

zitongzhan:
Ah there's a test failed on ci. Isn't the scheduler test.
Local testings were still fine
image

zitongzhan (approved):

PR #201, opened 2023-03-20
Update README.md

zitongzhan:
Syntax should be specified as doctest instead of python.
After the change, the cell is runable in Jetbrains doctest

Before:
image

After:
image


wang-chen:
It seems that Github doesn't support `doctest` rendering. Will close this PR temporality and reopen it in the future if GitHub changed their settings.

PR #208, opened 2023-03-31
add runtime torch version checking

zitongzhan:


pypose/__init__.py
```diff
@@ -14,3 +14,24 @@
...
+        if x.isdigit():
+            digits.append(int(x))
+        elif x.find('rc') != -1:
+            patch_version = x.split('rc')
+            digits.append(int(patch_version[0]) - 1)
+            digits.append(int(patch_version[1]))
+    return digits
+
+
+torch_min_ver = '2.0'
```
wang-chen, line 31: this will need change once torch version is upgraded, better to read `requirements/runtime.txt`

pypose/__init__.py
```diff
@@ -14,3 +14,24 @@
...
+torch_version = digit_version(torch.__version__)
+
+
+assert digit_version(torch_min_ver) <= torch_version, \
```
wang-chen, line 35: How about 
```
from packaging import version
assert version.parse('2.0') <= version.parse(torch.__version__)
```

wang-chen (changes requested):

pypose/__init__.py
zitongzhan: that works

zitongzhan: Also add `packaging` to runtime.txt?

wang-chen: may not be necessary, our existing dependencies need it?

pypose/__init__.py
zitongzhan: `requirements/runtime.txt` is outside the package, so it is hard to do an i/o read from it during runtime.

pypose/__init__.py
zitongzhan: Indeed not necessary, checked that the commit afterwards passes testing

pypose/__init__.py
```diff
@@ -14,3 +16,9 @@
 from .lietensor.function import *
 from . import module, optim
 from .basics import *
+
+
+torch_min_ver = '2.0'
```
wang-chen, line 21: then delete line 21

wang-chen (approved):

PR #214, opened 2023-04-06
Dmpc non-linear

hakunanh:
add nonlinear lqr and mpc


wang-chen:
1. Don't have to define `NLQR`, we can extend `LQR` to accept optional args, e.g., `states`, and default to `None`.
2. Put the loop of set_refpoint and get A,B into loop of forward dynamics, one loop is enough to to implement.

Issue #187, opened 2023-03-02
Distance between rotations


HongLouyemeng:
### 🚀 The feature, motivation and pitch

For two rotations, the distance is a very important property

### Alternatives

Here are some reference links
1.http://www.boris-belousov.net/2016/12/01/quat-dist/
2.https://www.cs.cmu.edu/~cga/dynopt/readings/Rmetric.pdf
3.https://lcvmwww.epfl.ch/publications/data/articles/63/Means_and_Averaging_in_the_Group_of_Rotations.pdf

wang-chen:
@ShaoshuSu

ShaoshuSu:
Hi @HongLouyemeng, thanks for your suggestion : ) 
We have noticed that topic and will work on that soon

Murphy41:
The feature has been partly implemented in Pull Request #216 (in posediff function).

PR #216, opened 2023-04-12
ICP

XiangyuChen0212:
Support batched input


ICP.py
```diff
@@ -0,0 +1,109 @@
...
+        nn = dist.topk(k, largest=False)
+        return nn
+
+
+if __name__=="__main__":
+    input_pc = torch.randn([10, 20, 3])
+    transT = 0.1 * torch.randn([10, 1, 3])
+    print("The ori transT is", transT)
+    transR = pp.randn_SO3(10)
+    print("The ori transR is", transR)
```
zitongzhan, line 102: How about a random SE3?

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+    r'''
+    Iterative Closest Point (ICP) using Singular Value Decomposition (SVD).
+
+    Args:
+        max_iterations: the max iteration number
+        tolerance: the tolerance of error for early stopping
+        is_matching: whether the input points set and the target points set have been matched
+    '''
+
+    def __init__(self, max_iterations=200, tolerance=0.0001, is_matching=False):
```
zitongzhan, line 16: we want to avoid usage of `_`, can change var name `is_matching` -> `matched`

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+    transT = 0.1 * torch.randn([10, 1, 3])
+    print("The ori transT is", transT)
+    transR = pp.randn_SO3(10)
+    print("The ori transR is", transR)
+    transR = transR.matrix()
+
+    output_pc = (input_pc @ transR) + transT
+
+    icpsvd = ICP(is_matching=False)
+    result = icpsvd.forward(input_pc, output_pc)
```
zitongzhan, line 108: `result = icpsvd(input_pc, output_pc)` is the standard PyTorch way of function call

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+        '''
+        dif = torch.stack([p2[i].unsqueeze(-2) - p1[i]
+                           for i in range(p1.shape[0])])
+
+        dist = torch.norm(dif, dim=-1)
+        nn = dist.topk(k, largest=False)
+        return nn
+
+
+if __name__=="__main__":
```
zitongzhan, line 97: This trial run shall be moved to test

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            p1: input points set
+            p2: the matched target points set
+
+        Returns:
+            R: the rotation matrix between p1 and p2
+            T: the translation matrix between p1 and p2
+        '''
+        p1_centroid = p1.mean(-2).unsqueeze(-2)
```
zitongzhan, line 59: `p1_centroid = p1.mean(-2, keepdim=Ture)`

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+                    break
+                prev_err = mean_err
+
+            transR, transT = self.get_transform(p1, temp_pc)
+            transT = torch.transpose(transT, 1, 2)
+            T = torch.cat([transR, transT], dim=2)
+            return pp.mat2SE3(T, check=False)
+        else:
+            transR, transT = self.get_transform(p1, p2)
+            transT = torch.transpose(transT, 1, 2)
```
zitongzhan, line 44: need to use negative dimension indexing. If dim 1, 2 are referring to the last dimensions, you could use `transT.mT`

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+            T = torch.cat([transR, transT], dim=2)
+            return pp.mat2SE3(T)
+
+    def get_transform(self, p1: torch.tensor, p2: torch.tensor):
```
zitongzhan, line 48: There is an existing transform calculation module just merged to main branch. Consider using it so that we don't have to keep two copies of the same functionality
https://github.com/pypose/pypose/blob/60a38038cc6b18811749084b85808bde510357fb/pypose/module/pnp.py#L294

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+            tolerance: the threshold of min distance
+
+        Returns:
+            distances: the min distance between point in p1 and its nearest neighbor
+            indices: the index of the nearest neighbor point in p2
+        '''
+        dif = torch.stack([p2[i].unsqueeze(-2) - p1[i]
+                           for i in range(p1.shape[0])])
+
+        dist = torch.norm(dif, dim=-1)
```
zitongzhan, line 92: In PyPose, `dif.norm(dim=-1)` is considered to be more straightforward.

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+            p1: the source points set
+            p2: the target points set
+            tolerance: the threshold of min distance
+
+        Returns:
+            distances: the min distance between point in p1 and its nearest neighbor
+            indices: the index of the nearest neighbor point in p2
+        '''
+        dif = torch.stack([p2[i].unsqueeze(-2) - p1[i]
+                           for i in range(p1.shape[0])])
```
zitongzhan, line 90: The for loop could be avoided by using broadcasting.

zitongzhan (changes requested):

ICP.py
```diff
@@ -0,0 +1,109 @@
+import torch
+import pypose as pp
```
Murphy41, line 2: Here I have a quick question (maybe more for Zitong and Chen): From Rule 12 in contributor guideline 
"Don’t import pypose in pypose modules, only import pypose in tests and examples."
Should we use import the exact function or module of the pypose here? I reviewed a few other modules, and some of them avoided to import pypose, but some of them didn't.

zitongzhan: import the exact function.
A nice example would be importing homo2cart in pnp.py

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+        super().__init__()
+        self.max_iterations = max_iterations
+        self.tolerance = tolerance
+        self.is_matching = is_matching
+
+    def forward(self, p1, p2):
+        ori_p2 = p2.clone()
+        temp_pc = p1.clone()
+        iter_num = 0
+        prev_err = 0
```
Murphy41, line 26: As Zitong mentioned, the usage of `_` can be avoided. The variable name p1 and p2 can be improved to make it clearer. And 'ori_p2' can be simplified as 'original', 'originalpc' or 'existpc', etc. Given we only have pointcloud here, I think it's clear enough to use "original" and "temp", but if we want to make it clearer, 'existpc' could be better. Some other improvements can be done:
iter_num -> iteration
prev_err -> error

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+        if (not self.is_matching):
+            while iter_num <= self.max_iterations:
+                iter_num += 1
+                nn = self.nearest_neighbor(temp_pc, p2)
+                mean_err = nn.values.mean()
+                transR, transT = self.get_transform(temp_pc, ori_p2[:, nn.indices[-1], :].squeeze(-2))
```
Murphy41, line 32: Based on Rule 9 in the contributor guideline, probably use  state[..., n] instead of state[:, n]. (I'm not sure about whether I got the correct understanding here, as I'm not familiar with batch operations in PyTorch. If my suggestion is wrong here, please ignore.)

XiangyuChen0212: Ok, I will modify it.

zitongzhan: Yes. The same reason why we use negative indexing

zitongzhan:
Please commit your changes directly to the branch for PR. Each time you finished revising, you may re-request review.

ICP.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def forward(self, p1, p2):
+        ori_p2 = p2.clone()
+        temp_pc = p1.clone()
+        iter_num = 0
+        prev_err = 0
+        if (not self.is_matching):
+            while iter_num <= self.max_iterations:
+                iter_num += 1
+                nn = self.nearest_neighbor(temp_pc, p2)
```
Murphy41, line 30: `nn` is the module you imported from `torch`, it's better to change the variable name.

test/module/test_icp.py
wang-chen, line 1: follow the format of other test examples.

pypose/module/icp.py
```diff
@@ -0,0 +1,74 @@
...
+            indices: the index of the k nearest neighbor points in pc2
+        '''
+
+        diff = pc1.unsqueeze(-2) - pc2.unsqueeze(-3)
+        if norm == 1:
+            distance = torch.sum(torch.abs(diff), dim=-1)
+        elif norm == 2:
+            distance = torch.sqrt(torch.sum(diff ** 2, dim=-1))
+        else:
+            raise ValueError("Invalid norm. Only 1-norm and 2-norm are supported.")
```
wang-chen, line 64: How about `distance = diff.norm(dim=-1, ord=norm)`

pypose/module/icp.py
```diff
@@ -0,0 +1,74 @@
...
+        temppc = p1.clone()
+        iter = 0
+        err = None
+        if (self.init_transform != None):
+            temppc = self.init_transform.Act(temppc)
+        while iter <= self.steplim:
+            iter += 1
+            neighbors = self._k_nearest_neighbor(temppc, p2)
+            errnew = torch.mean(neighbors.values, dim=-1)
+            T = EPnP._points_transform(temppc, p2[:, neighbors.indices[-1],:].squeeze(-2))
```
wang-chen, line 30: I suggest defining an independent function for `points_transform` and expose to users, then both `EPnP`and `ICP` call this new function.

pypose/module/icp.py
```diff
@@ -0,0 +1,74 @@
...
+from . import EPnP
+from .. import LieTensor
+class ICP(torch.nn.Module):
+    r'''
+    Iterative Closest Point (ICP) using Singular Value Decomposition (SVD).
+    Args:
+        steplim: the max step number
+        tol: the tolerance of error for early stopping
+        matched: whether the input points set and the target points set have been matched
+    '''
```
wang-chen, line 11: add doc rst definition, refer to [here](https://github.com/pypose/pypose/blob/main/CONTRIBUTING.md#12-writing-documentation)

pypose/module/icp.py
zitongzhan: Should move `points_transform` to `pypose.geometry`?

wang-chen: yes, we can do that

test/module/test_icp.py
```diff
@@ -32,31 +32,30 @@ def l_shape_wall(b, side1, side2, resolution=0.01, noise_std=0.01):
...
         pc2 = tf.unsqueeze(-2).Act(pc2)
 
         return pc1, pc2, tf
 
     def random_point_cloud(b, num_points):
         pc1 = torch.rand(b, num_points, 3)
         tf = pp.randn_SE3(b)
-        print("The true tf is", tf)
         pc2 = tf.unsqueeze(-2).Act(pc1)
         return pc1, pc2, tf
```
zitongzhan, line 43: You could write the following test:
1. obtain a point cloud from somewhere
2. Keep two copies of the point cloud, randomly permute one of them and then apply random *known* transformations
3. Input the two point clouds to ICP, and use ICP to recover the transformation. Compare the result with the known transformation.

pypose/module/icp.py
Murphy41: > Should move `points_transform` to `pypose.geometry`?

Hi Zitong, will the points_transformation function be moved to the geometry module in this pull request? If so, would you prefer to make the change yourself or would you like our assistance with the modification?

zitongzhan: You may move that function directly in this PR

pypose/function/geometry.py
```diff
@@ -159,3 +161,122 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
            intrinsics.size(-1) == intrinsics.size(-2) == 3, "Shape not compatible."
     img_repj = point2pixel(points, intrinsics, extrinsics)
     return (img_repj - pixels).norm(dim=-1)
+
+
+def knn(pc1, pc2, k = 1, norm = 2):
```
zitongzhan, line 166: Don’t use spaces around the = sign when used to indicate a keyword argument
See https://peps.python.org/pep-0008/#other-recommendations

tests/module/test_icp.py
```diff
@@ -0,0 +1,72 @@
...
+    def load_point_cloud(self):
+        download_and_extract_archive('https://github.com/Murphy41/laser-scan-pt/' \
+                                     'releases/download/v0.0/icp-test-data.pt.zip',\
+                                    './tests/module')
+        loaded_tensors = torch.load('./tests/module/icp-test-data.pt')
+        pc1 = loaded_tensors['pc1'].squeeze(-3)
+        pc2 = loaded_tensors['pc2'].squeeze(-3)
+        return pc1, pc2
+
+    def test_icp_laserscan(self):
```
zitongzhan, line 16: Formatting
See https://peps.python.org/pep-0008/#indentation

tests/module/test_icp.py
```diff
@@ -0,0 +1,72 @@
...
+        gt_rot = self.tf_gt.rotation().matrix()
+        gt_t = self.tf_gt.translation()
+
+        def rmse_rot(pred, gt):
+            return torch.linalg.norm((pred - gt), dim=-1, ord=2).mean()
+
+        def rmse_t(pred, gt):
+            return torch.linalg.norm((pred - gt), dim=-1, ord=2).mean()
+
+        print("Pypose ICP solution, rmse of R:", rmse_rot(rot, gt_rot))
```
zitongzhan, line 63: Use this function to make it an actual test https://pytorch.org/docs/stable/testing.html

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
...
+        temppc = psrc.clone()
+        iter = 0
+        err = 0
+        dim = psrc.shape
+        while iter <= self.steplim:
+            iter += 1
+            neighbors = knn(temppc, ptgt)
+            knndist = neighbors.values.squeeze(-1)
+            knnidx = neighbors.indices
+            errnew = torch.mean(knndist, dim=-1)
```
zitongzhan, line 92: `knndist.mean(dim=-1)` would have better readability

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
...
+            psrc(``torch.Tensor``): The source point cloud tensor with
+                [..., points_num, 3] shape.
+            ptgt(``torch.Tensor``): The target point cloud tensor with
+                [..., points_num, 3] shape.
+
+        Returns:
+            ``LieTensor``: The estimated transformation (``SE3type``) from psrc to ptgt.
+
+        '''
+        temppc = psrc.clone()
```
zitongzhan, line 83: Is this statement needed? `temppc = psrc.clone()`

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
...
+        >>> pc2 = torch.tensor([[[0.2, 0.1, 0.],
+        ...                      [1.1397, 0.442, 0.],
+        ...                      [2.0794, 0.7840, 0.]]])
+        >>> icp = pp.module.ICP()
+        >>> icp(pc1, pc2)
+        SE3Type LieTensor:
+        LieTensor([[0.2000, 0.1000, 0.0000, 0.0000, 0.0000, 0.1736, 0.9848]])
+
+    '''
+    def __init__(self, steplim = 200, tol = 1e-6, tf = None):
```
zitongzhan, line 61: the same space usage suggestion

tests/module/test_icp.py
```diff
@@ -0,0 +1,72 @@
...
+        b2 = 2
+        num_points = 2
+        pc1 = torch.rand(b1, b2, num_points, 3)
+        self.tf_gt = pp.randn_SE3(b1,b2)
+        pc2 = self.tf_gt.unsqueeze(-2).Act(pc1)
+        icp = pp.module.ICP()
+        self.result = icp(pc1, pc2)
+        self.rmse_results()
+
+    def rmse_results(self):
```
zitongzhan, line 51: better to let prediction `result` and `tf_gt` being input of the function. and return an error.

tests/module/test_icp.py
```diff
@@ -0,0 +1,72 @@
...
+        rot = self.result.rotation().matrix()
+        t = self.result.translation()
+        gt_rot = self.tf_gt.rotation().matrix()
+        gt_t = self.tf_gt.translation()
+
+        def rmse_rot(pred, gt):
```
zitongzhan, line 57: @wang-chen What's the best package to place this function in? It is a commonly used one. I am thinking of `geometry`

zitongzhan (changes requested):

zitongzhan:
Nice work done. See comments above.

pypose/module/icp.py
XiangyuChen0212: Yes, in the end, the origin psrc is used

tests/module/test_icp.py
Murphy41: Hi Zitong, is this comment for line 10 or other lines?

tests/module/test_icp.py
Murphy41: Certainly, but this approach may only be effective for static input as opposed to random input, since ICP can merely achieve local optima and might converge to a local minimum, causing a failure in the tests. I anticipate these tests are intended to assess functionality, so I can modify them to utilize static input point clouds and transformations. Consequently, any negative results would highlight the function's incorrectness rather than the inherent limitations of the algorithms.

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
+import torch
+from .. import lietensor
+from .. import knn, svdtf
+
+class ICP(torch.nn.Module):
+    r'''
+    This class implements the batched Iterative Closest Point (ICP) algorithm to find a
+    similarity transformation ( :math:`T` ) between two sets of 3-dimensional points
+    using Singular Value Decomposition (SVD). It's important to note that the solution
+    found is only a local optimum.
```
wang-chen, line 10: Batched Iterative Closest Point (ICP) algorithm to find a rigid transformation between two sets of points using Singular Value Decomposition (SVD).

add warning section for 'It's important to note that the solution
    found is only a local optimum.' First sentence of the doc only explains its purpose.

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
...
+
+    The algorithm takes two input point clouds: source point cloud (psrc) and target
+    point cloud (ptgt). The objective is to find the optimal similarity transformation
+    (:math:`T`) to minimize the error between the transformed source point cloud and the
+    target point cloud as shown in the equation:
+
+    .. math::
+        \begin{align*}
+            \underset{T}{\operatorname{arg\,min}} \sum_i \|ptgt_j - T \cdot psrc_i\|,
+        \end{align*}
```
wang-chen, line 28: We don't use string such as 'ptgt' in math variable name. use normal fonts in super- and subscript, not italic. see generated doc.

pypose/module/icp.py
```diff
@@ -0,0 +1,100 @@
...
+            psrc(``torch.Tensor``): The source point cloud tensor with
+                [..., points_num, 3] shape.
+            ptgt(``torch.Tensor``): The target point cloud tensor with
+                [..., points_num, 3] shape.
+
+        Returns:
+            ``LieTensor``: The estimated transformation (``SE3type``) from psrc to ptgt.
+
+        '''
+        temppc = psrc.clone()
```
wang-chen, line 83: it seems that we don't need to clone?

zitongzhan:
Docs preview: https://pypose--216.org.readthedocs.build/en/216/generated/pypose.module.ICP.html

pypose/function/geometry.py
```diff
@@ -159,3 +161,122 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
...
+def svdtf(pc1, pc2):
+    r'''
+    Compute the rigid similarity transformation (rotation and translation) between two
+    sets of points using Singular Value Decomposition (SVD).
+
+    Args:
+        pc1 (``torch.Tensor``): The coordinates of the first set of points.
+            The shape has to be (..., N1, 3).
+        pc2 (``torch.Tensor``): The coordinates of the second set of points.
+            The shape has to be (..., N2, 3).
```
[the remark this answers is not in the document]
Murphy41: > it seems that this function requires the point clouds have to be associated and thus the number of points has to be the same? so no need `N1`, `N2`, use `N` instead.

I'm not entirely certain, but based on my testing, the code appears to work when N1 is equal to 1 and N2 has different values. My understanding is that N1 and N2 must either be equal or one of them should be 1, and the same requirement applies to batch numbers. This constraint arises from the torch.matmul calculation in bvv, as it requires a size match for non-singleton dimensions. Are we planning to add an assert function to enforce size matching? This approach seems reasonable for the ICP or EPnP case, but I'm unsure if it's a good practice for a more generalized function.

pypose/function/geometry.py
```diff
@@ -159,3 +161,134 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
            intrinsics.size(-1) == intrinsics.size(-2) == 3, "Shape not compatible."
     img_repj = point2pixel(points, intrinsics, extrinsics)
     return (img_repj - pixels).norm(dim=-1)
+
+
+def knn(pc1, pc2, ord=2, k=1, dim=-1, largest=False, sorted=True):
```
Murphy41, line 166: image
Would anyone be able to help in resolving the annoying spacing issue present in `dim=-1`?

pypose/function/geometry.py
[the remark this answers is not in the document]
Murphy41: @zitongzhan Hi Zitong, how do you think about the dimension of N1 and N2 in this function?

pypose/function/geometry.py
```diff
@@ -177,7 +177,7 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
     return (img_repj - pixels).norm(dim=-1)
 
 
-def knn(pc1, pc2, ord=2, k=1, dim=-1, largest=False, sorted=True):
+def knn(pc1, pc2, dim=-1, ord=2, k=1, largest=False, sorted=True):
```
Murphy41, line 180: image
Would anyone be able to help in resolving the annoying spacing issue present in `dim=-1`?

pypose/function/geometry.py
```diff
@@ -159,3 +175,210 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
...
+            The shape has to be (..., N, 3).
+        pc2 (``torch.Tensor``): The coordinates of the second set of points.
+            The shape has to be (..., N, 3).
+
+    Returns:
+        ``LieTensor``: The rigid transformation matrix in ``SE3Type``  that
+        minimizes the mean squared error between the input point sets.
+
+    Warning:
+        The number of points N has to be the same for both point clouds.
```
wang-chen, line 284: suggest: 'The input point clouds have to be associated and have the same number of points N.'

pypose/function/geometry.py
```diff
@@ -159,3 +175,210 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
...
+    U, S, Vh = torch.linalg.svd(M)
+    R = U @ Vh
+    mask = (R.det() + 1).abs() < 1e-6
+    R[mask] = - R[mask]
+    t = pc2ctn.mT - R @ pc1ctn.mT
+    T = torch.cat((R, t), dim=-1)
+    return mat2SE3(T, check=False)
+
+
+def posediff(ref, est, aggregate=False, mode=1):
```
wang-chen, line 313: it seems that this function is not used in any functions for now. Here is my concern on the API:
- The mode 0 contains quaternion, which should be a LieTensor, as users may want to use its properties for further computation.
- How will `posediff` be used is still not clear, I suggest not to expose to users for now, as once release we will need to maintain this API and this will pose issues if we find a better API in the future.

pypose/module/icp.py
```diff
@@ -0,0 +1,97 @@
...
+    '''
+    def __init__(self, steps=200, tol=1e-6, init=None):
+        super().__init__()
+        self.steps = steps
+        self.tol = tol
+        self.init = init
+        if init != None:
+            assert is_SE3(init), "The input initial transformation is not SE3Type."
+
+    def forward(self, psrc, ptgt):
```
wang-chen: suggest to change args name to `source` and `target`, as the class name already says it is for point cloud.

wang-chen: suggest add optional args for `ord`, `dim` to pass to `knn`.

pypose/module/icp.py
```diff
@@ -0,0 +1,97 @@
...
+        It's important to note that the solution found is only a local optimum.
+
+    '''
+    def __init__(self, steps=200, tol=1e-6, init=None):
+        super().__init__()
+        self.steps = steps
+        self.tol = tol
+        self.init = init
+        if init != None:
+            assert is_SE3(init), "The input initial transformation is not SE3Type."
```
wang-chen, line 67: `init` is not used in implementation, do you want to use it to transform the 'temppc' before entering the loop?

wang-chen (changes requested):

pypose/function/geometry.py
wang-chen: yes, please.

Murphy41: Will do after merge :)

pypose/function/geometry.py
```diff
@@ -159,3 +175,210 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
...
+        largest (``bool``, optional): Return the k nearest or furthest neighbors. If
+            ``largest`` is set to ``True``, then the k furthest neighbors are returned.
+            Default: ``False``.
+        sorted (``bool``, optional): Return the sorted or unsorted k nearest neighbors. If
+            ``sorted`` is set to ``True``, it will make sure that the returned k nearest
+            neighbors are themselves sorted. Default: ``True``.
+
+    Note:
+        If ``sorted`` is set to ``False``, the output will be unordered and not
+        necessarily aligned with the index of the input point cloud.
```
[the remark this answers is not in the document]
Murphy41: Sorry, my expression is misleading. Will update. The note I want to write here is to highlight the output will be unspecified and not necessary sorted along the index (For more details, please refer to https://github.com/pytorch/pytorch/issues/3982). And yes, the index is aligned between `topk.values` and `topk.indices`.

pypose/function/geometry.py
Murphy41: Agree. This function is currently used in test_icp and potentially will be used in other tests. The initial idea is to calculate the error between two transformations, where the translational error is easy to define, but the rotational ones are a bit tricky (like mentioned in https://github.com/pypose/pypose/issues/187). This function is not clearly structured with limited functionality at current stage. I think we can make it an internal function for now, and have another pull request dedicate to it.

tests/module/test_icp.py
```diff
@@ -0,0 +1,63 @@
...
+import pypose as pp
+from torchvision.datasets.utils import download_and_extract_archive
+
+class TestICP:
+
+    def test_icp_laserscan_data(self):
+        # pc1 and pc2 has different numbers of points
+        download_and_extract_archive('https://github.com/Murphy41/laser-scan-pt/' \
+                                     'releases/download/v0.0/icp-test-data.pt.zip',\
+                                     './tests/module')
```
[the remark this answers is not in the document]
Murphy41: Thanks :+1:

pypose/module/icp.py
Murphy41: > suggest add optional args for `ord`, `dim` to pass to `knn`.

Yes, we can pass the 'ord' and 'dim' arguments to the KNN function, but I am slightly concerned that this may confuse users.  Retaining these arguments in the KNN function appears to be a reasonable decision for generalization purposes, as other modules or functions might necessitate greater flexibility. However, specifically for the ICP or PNP modules, since the point cloud is calculated using the SE3 Lie algebra and the point cloud format is defined as (..., N, 3), the 'dim' value must be -1 in this case. Additionally, 'ord' is naturally set to 2, as ICP involves 3D spatial geometry. these arguments are optional and should not cause significant concern at this stage. We can include them for now and, if required, make refinements in the future.

Murphy41: > how about we also change the point cloud args of `knn` and `svdtf` to `source` and `target` as well?

Thanks for suggestion. I changed the variable names for svdtf, as it has the flow direction (find tf from one to the other). For knn, i'm slightly concerned that `source` and `target` may confused users, I would suggest to keep pc1 and pc2.

pypose/function/geometry.py
```diff
@@ -159,3 +175,210 @@ def reprojerr(points, pixels, intrinsics, extrinsics=None):
...
+        >>> pp.svdtf(pc1, pc2)
+        SE3Type LieTensor:
+        LieTensor([1., 1., 1., 0., 0., 0., 1.])
+    '''
+    assert pc1.size(-2) == pc2.size(-2), "The number of the points N has to be the same."
+    pc1ctn = pc1.mean(dim=-2, keepdim=True)
+    pc2ctn = pc2.mean(dim=-2, keepdim=True)
+    pc1t = pc1 - pc1ctn
+    pc2t = pc2 - pc2ctn
+    M = bvv(pc2t, pc1t).sum(dim=-3)
```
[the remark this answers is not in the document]
Murphy41: Thank you for your suggestion. I am not familiar with this function, and will investigate it over the coming days.

wang-chen (changes requested):
Just tested, this module has bug given high batch broadcastable source and target shape, e.g., source`(2, 100, 3)` target `(100, 3)` or, source`(2, 1, 100, 3)` target `(2, 2, 100, 3)`.

Murphy41:
> Just tested, this module has bug given high batch broadcastable source and target shape, e.g., source`(2, 100, 3)` target `(100, 3)`.

Hi Chen, I've fixed the bug :smile:

pypose/module/icp.py
```diff
@@ -0,0 +1,102 @@
...
+            temporal = init.unsqueeze(-2) @ temporal
+        for _ in range(self.steps):
+            knndist, knnidx = knn(temporal, target, k=1, ord=ord, dim=dim)
+            errnew = knndist.squeeze(-1).mean(dim=-1)
+            if torch.all((errnew - errlast).abs() < self.tol):
+                break
+            errlast = errnew
+            target = target.view(*([1] * (knnidx.dim() - target.dim())), *target.shape)
+            target = target.expand(*knnidx.shape[:-2], target.size(-2),target.size(-1))
+            knntarget = torch.gather(target, -2, knnidx.expand(*knnidx.shape[:-1],
```
wang-chen, line 98: How about out of the loop, we define
```python
batch = torch.broadcast_shapes(source.shape[:-2], target.shape[:-2])
```
then replace these two line by
```python
target = target.expand(batch + target.shape[-2:])
knntarget = torch.gather(target, -2, knnidx.expand(batch + source.shape[-2:]))
```

Murphy41: It looks much more elegant now!

wang-chen (approved):

Murphy41 (approved):

wang-chen (approved):

PR #217, opened 2023-04-12
Dmpc nonlinear new version

hakunanh:
update lqr


zitongzhan:
A preview of doc: https://pypose--217.org.readthedocs.build/en/217/generated/pypose.module.LQR.html

pypose/module/lqr.py
```diff
@@ -88,19 +88,104 @@ class LQR(nn.Module):
             \mathbf{c} \left( \mathbf{\tau}_t \right) = \frac{1}{2}
             \mathbf{\tau}_t^\top\mathbf{Q}_t\mathbf{\tau}_t + \mathbf{p}_t^\top\mathbf{\tau}_t
 
+    A discrete-time non-linear system can be described as:
```
aerogjy, line 91: Let's add some transition here. Maybe a new paragraph title? or some bold text. For the nonlinear system, sometimes people want to solve MPC problem with iterative LQR. Add the citation of this paper https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=dfc73c664f85562956e4b98b3102ebd7e41906ae

pypose/module/lqr.py
```diff
@@ -158,41 +249,53 @@ def __init__(self, system, Q, p, T):
         assert self.Q.device == self.p.device, "device not compatible."
         assert self.Q.dtype == self.p.dtype, "tensor data type not compatible."
 
-    def forward(self, x_init):
+    def forward(self, x_init, current_x, current_u, time):
```
aerogjy, line 252: Let's do not add these to the argument. These should be generated inside the class rather than let's the user provide. Because these are some intermediate variable you need to used for the convenience of the algorithm. Not necessary for using this class. See the original mpc.pytorch code.

pypose/module/lqr.py
```diff
@@ -158,41 +249,53 @@ def __init__(self, system, Q, p, T):
...
+        K, k = self.lqr_backward(current_x, current_u, time)
+        x, u, cost = self.lqr_forward(x_init, current_x, current_u, K, k)
         return x, u, cost
 
-    def lqr_backward(self):
+    def lqr_backward(self, current_x, current_u, time):
+
+        # Q: (n_batch*, T, N, N), p: (n_batch*, T, N), where n_batch* can be any batch dimensions, e.g., (2, 3)
+        n_batch = self.p.shape[:-2]
+        ns, nc = current_x.size(-1), current_u.size(-1)
```
aerogjy, line 278: I noticed these appeared multiple time. Why don't you just obtain these once at the beginning of the __init_ to make them the class member. Because one the system is passed through the class, ns, nc should be fixed. There is no need to compute them several times.

pypose/module/mpc.py
```diff
@@ -0,0 +1,49 @@
...
+                    'u': u,
+                    'cost': cost,}
+            else:
+                if cost <= best['cost']+ self.eps:
+                    best['x'] = x
+                    best['u']= u
+                    best['cost'] = cost
+
+        if self.step > 1:
+            current_x = best['x']
```
aerogjy, line 43: This if should be inside the for loop. I was surprised how this thing can work if you put them here. Because you need to update current_x, current_u for next iteration to do linearization.

pypose/module/mpc.py
```diff
@@ -0,0 +1,49 @@
...
+class MPC(nn.Module):
+
+    def __init__(self, system, T, step=10, eps=1e-4):
+        super().__init__()
+        self.system = system
+        self.T = T
+        self.step = step
+        self.eps = eps
+
+    def forward(self, Q, p, x_init, current_x, current_u, time):
```
aerogjy, line 15: Like I said, do not put current_x, current_u here. You should generate the initial sequence inside the class rather than let the user generate the non-sense sequences by themselves. See the original mpc.pytorch code.

Also, in the original code, the author always only use the u sequence from last iteration to re-generate x sequence. Should we also do that or we keep using the x sequence from last iteration? Let's double check the math.

pypose/module/mpc.py
```diff
@@ -0,0 +1,49 @@
...
+
+        for i in range(self.step):
+            current_u.requires_grad = False
+            if not current_u.requires_grad:
+                current_u = current_u
+            else:
+                current_u = current_u.detach()
+
+            lqr = pp.module.LQR(self.system, Q, p, self.T)
+            x, u, cost = lqr(x_init, x, u, time)
```
aerogjy, line 28: You are mixing using self. T and time. Why don't we only use one of them. My understanding is they are the same? Just the fixed time horizon?

examples/module/mpc/cartpole.py
```diff
@@ -0,0 +1,153 @@
...
+                     torch.mean((m_cart - expert['m_cart'])**2) + \
+                     torch.mean((m_pole - expert['m_pole'])**2)
+
+        loss_f.write('{},{}\n'.format(traj_loss.item(), model_loss.item()))
+        loss_f.flush()
+
+        plot_interval = 1
+        if i % plot_interval == 0:
+            os.system('./plot.py "{}" &'.format(args.save))
+            print(len, expert['len'])
```
aerogjy, line 145: Can we print some info which are easy for the user? Let's say these length and masses of expert and agent so that user can easily see what are you printing.

examples/module/mpc/linear.py
```diff
@@ -0,0 +1,140 @@
...
+                print(B, expert['B'])
+
+            print('{:04d}: traj_loss: {:.4f} model_loss: {:.4f}'.format(
+                i, traj_loss.item(), model_loss.item()))
+            print("backward_time = ", backward_time)
+            print("overall_time = ", overall_time)
+
+
+if __name__=='__main__':
+    train = TrainMPC()
```
aerogjy, line 138: This example doesn't seem to be correct when I run it. The model loss keep increasing.

test/module/test_lqr.py
```diff
@@ -135,21 +145,117 @@ def B(self):
...
+        torch.testing.assert_close(x_ref, x, atol=1e-5, rtol=1e-3)
+        torch.testing.assert_close(u_ref, u, atol=1e-5, rtol=1e-3)
+
 
 if __name__ == '__main__':
     test = TestLQR()
     device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
     test.test_lqr_linear(device)
     test.test_lqr_ltv(device)
+    test.test_nlqr_cartpole(device)
```
aerogjy, line 261: Let's provide a separate test_mpc.py.

pypose/module/lqr.py
```diff
@@ -158,41 +249,53 @@ def __init__(self, system, Q, p, T):
...
-                F = torch.cat((self.system.A, self.system.B), dim=-1)
+                self.system.set_refpoint(state=current_x[...,t,:], input=current_u[...,t,:], t=time[t])
+                A = self.system.A.squeeze(-2)
+                B = self.system.B.squeeze(-1)
+                F = torch.cat((A, B), dim=-1)
                 Qt = self.Q[...,t,:,:] + F.mT @ V @ F
-                qt = self.p[...,t,:] + bmv(F.mT, v)
-                if self.system.c1 is not None:
-                    qt = qt + bmv(F.mT @ V, self.system.c1)
+                qt = p_new[...,t,:] + bmv(F.mT, v)
```
aerogjy, line 298: How can this be generalize to the linear system?

aerogjy (changes requested):
First round of review. 
1. The documentation of MPC is missing. No test file for MPC. No MPC in the doc/module.
2. the API design shouldn't have current_x, current_u, those should be generated inside the class rather than let the user provide.
3. The example linear.py doesn't seem correct. The model loss keep increasing.
4. The implementation of iterative lqr in mpc.py is not correct. The sequence update should be inside the for loop for next iteration linearization.
5. Some other coding style issues I've commented in the specific place.

examples/module/mpc/cartpole.py
```diff
@@ -0,0 +1,146 @@
...
+    fname = os.path.join(args.save, 'cartpole losses.csv')
+    loss_f = open(fname, 'w')
+    loss_f.write('im_loss,mse\n')
+    loss_f.flush()
+
+    mpc_expert = pp.module.MPC(expert_cartPoleSolver, T, step=15).to(device)
+    x_true, u_true, cost_true = mpc_expert(expert['Q'], expert['p'], x_init, dt, current_u)
+
+    def get_loss(_len, _m_cart, _m_pole):
+        x_init = torch.tensor([[0, 0, torch.pi, 0]], device=device)
```
aerogjy, line 107: Why are you using a fixed x_init? There will be no meaning to do 1000 iteration....

examples/module/mpc/cartpole.py
```diff
@@ -0,0 +1,146 @@
+import os
+import torch
+import shutil
+import argparse
+import setproctitle
+import pypose as pp
+import pickle as pkl
+import torch.optim as optim
+
+def main():
```
aerogjy, line 10: I run this example. It converges really slow and cannot converge to the gt values, still significant amount of error.

aerogjy (changes requested):
Linear example works. Cartpole example is really slow and cannot converge to the gt value. One potential bug is in the x_init.
The MPC doc needs a big update to make it clear. I would hand it over to @wang-chen after this round. But before that, I would like to double check the doc after the change.

aerogjy:
I realize we haven't tried the non-quardratic cost scenario. Perhaps, let's merge this first. Then, add that functionality.

wang-chen (approved):
temporarily approved for CVPR

wang-chen:
A list for change in a future formal release
image

PR #225, opened 2023-04-17
add an option for ``extrinsics`` to be ``None`` in ``reprojerr``

wang-chen:
add an option for ``extrinsics`` to be ``None`` in ``reprojerr``


zitongzhan (approved):
LGTM

PR #230, opened 2023-04-24
test2tests

zitongzhan:
We decided to change the name of `test` directory name to `tests`, which is more formal.


zitongzhan:
Verified on codespace that pytest works out-of-the-box

wang-chen (approved):

PR #232, opened 2023-04-28
simplify coo2hybrid & test Modularization

zitongzhan:
Contains the following changes
- in the logic simplification commit, placeholder was removed. This result in a non-trivial speedup once again. 
- test factory is up there. Through a single function, it is able to test multiple operators.

### DIFF
--- a/pypose/sparse/sparse_block_tensor.py
+++ b/pypose/sparse/sparse_block_tensor.py
@@ -80,7 +80,8 @@ def make_coo_indices_and_dims_from_hybrid(hybrid):
 
     # Shift the original indices.
     indices_ori = hybrid.indices() # (2, n_block)
-    indices_rp = indices_ori.unsqueeze(-1).repeat(1, 1, n_block_elem) #(2,n_block,n_block_elem)
+    indices_rp = indices_ori.unsqueeze(-1).expand(-1, -1, n_block_elem) #(2,n_block,n_block_elem)
+    # change to expand saves 0.3s for 10000 runs.
 
     # Compute the new indices by multiplying the block shape and adding the index shift.
     index_scale = torch.tensor(list(b_dim), dtype=torch.int64, device=hybrid.device)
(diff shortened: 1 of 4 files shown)


zitongzhan:
The above commit is to alleviate the problem caused by existence of non-dense blocks. This new commit also makes coo2hybrid's execution 100% faster.

pypose/sparse/sparse_block_tensor.py
```diff
@@ -188,32 +189,25 @@ def coo_2_hybrid(coo, proxy):
...
-    # Compute the indices of every element of coo inside their own respective blocks.
-    b_dim_t = torch.tensor(b_dim).to(dtype=torch.int64, device=coo.device).view((2, 1))
+    # in-block indices.
+    b_dim_t = torch.tensor(b_dim, dtype=torch.int64, device=coo.device).unsqueeze(-1)
     in_block_indices = coo.indices() % b_dim_t
+    # block indices.
+    block_indices = coo.indices() // b_dim_t
+    block_seq = torch.sparse_coo_tensor(
+            proxy.indices(), torch.arange(proxy.values().shape[0]), size=proxy.shape
+        ).to_dense()  # dense is fast in indexing
```
zitongzhan, line 200: The usage of `dense` here is for runtime concern. Indexing operation cost at least log(n) complexity for sparse tensor, but constant time for dense tensor. Here I prefer faster time over memory consumption.

huyaoyu (commented):
Please see the comments. Maybe let's setup a call sometime later.

pypose/sparse/sparse_block_tensor.py
zitongzhan: Sounds good. Can check the shape of proxy and decide whether to convert it to dense.

zitongzhan: Resolved in commit.

huyaoyu (approved):
Looks good. I might need to get more insight about the design of `coo_2_hybrid` when I review the high dimensional version.

PR #233, opened 2023-05-02
Scheduler

wang-chen:
Add a scheduler for ICP and other algorithms need a loop stop condition.

### DIFF
--- a/pypose/module/icp.py
+++ b/pypose/module/icp.py
@@ -1,5 +1,7 @@
 import torch
 from .. import knn, svdtf, is_SE3
+from .scheduler import ReduceToPlateau
+
 
 class ICP(torch.nn.Module):
     r'''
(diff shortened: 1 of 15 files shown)


wang-chen:
The current implementation is `pypose.module.ReduceToPlateau`, should we move to `pypose.utils.ReduceToPlateau`? Note that this is different from `pypose.optim.scheduler.StopOnPlateau` which can only work for optimizers, this can work for any algorithms that need a loop.

pypose/module/icp.py
```diff
@@ -49,18 +49,25 @@ class ICP(torch.nn.Module):
         >>> target = torch.tensor([[[0.2,      0.1,  0.],
         ...                         [1.1397, 0.442,  0.],
         ...                         [2.0794, 0.7840, 0.]]])
+        >>> scheduler = pp.module.ReduceToPlateau(steps=10, verbose=True)
         >>> icp = pp.module.ICP()
```
Murphy41, line 53: icp = pp.module.ICP(scheduler=scheduler)

Murphy41:
> The current implementation is `pypose.module.ReduceToPlateau`, should we move to `pypose.utils.ReduceToPlateau`? Note that this is different from `pypose.optim.scheduler.StopOnPlateau` which can only work for optimizers, this can work for any algorithms that need a loop.

1. Move `pypose.module.ReduceToPlateau` to `pypose.utils.ReduceToPlateau` sounds better to me, and it is a better practice to ensure modularity and prevent circular dependencies.
2. I had a quick look on `pypose.optim.scheduler.StopOnPlateau`, it appears to be an essential component of the optimizations. Are we going to merge these two schedulers? If that's not the case, what do you think about renaming the scheduler for non-optimization modules/functions to something else to prevent confusion?

PR #239, opened 2023-05-07
fix a bug in kernel.py

ShaoshuSu:
In the Tolerant class, both `self.a` and `self.b` are initialized as floats. This will cause a error when executing the forward method `part2 = (1 + torch.tensor(-self.a / self.b).exp())` line, as shown below:
```
    part2 = (1 + (-self.a / self.b).exp())
AttributeError: 'float' object has no attribute 'exp'
```
To resolve this issue, I have changed the data type of `-self.a / self.b` from float to torch.tensor. This should fix the bug.

### DIFF
--- a/pypose/optim/kernel.py
+++ b/pypose/optim/kernel.py
@@ -255,7 +255,7 @@ class Tolerant(nn.Module):
         '''
         assert torch.all(input >= 0), 'input has to be non-negative'
         part1 = (1 + ((input-self.a) / self.b).exp()).log()
-        part2 = (1 + torch.tensor(-self.a / self.b).exp())
+        part2 = (1 + torch.tensor(-self.a / self.b).exp()).log()
         return self.b * part1 - self.b * part2
 
 
@@ -296,4 +296,3 @@ class Scale(nn.Module):
             input (torch.Tensor): the input tensor (non-negative).
         '''
         return self.delta * input
-
--- a/pypose/optim/kernel.py
+++ b/pypose/optim/kernel.py
@@ -245,7 +245,6 @@ class Tolerant(nn.Module):
         super().__init__()
         assert a > 0, ValueError("a has to be positive: {}".format(a))
         assert b < 0, ValueError("b has to be negative: {}".format(b))
-        self.offset = b * math.log((1 + math.exp(-a / b)))
         self.a, self.b = a, b
 
     def forward(self, input: Tensor) -> Tensor:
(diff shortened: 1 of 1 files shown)


ShaoshuSu:
sorry there is another bug in the previous code, I will fix it later

wang-chen:
> sorry there is another bug in the previous code, I will fix it later

Could you also check whether the cited figure in doc is correct when fixing it? See the figure [in this doc](https://pypose.org/docs/main/generated/pypose.optim.kernel.Tolerant/).

ShaoshuSu:
> > sorry there is another bug in the previous code, I will fix it later
> 
> Could you also check whether the cited figure in doc is correct when fixing it? See the figure [in this doc](https://pypose.org/docs/main/generated/pypose.optim.kernel.Tolerant/).

Yes, sure

ShaoshuSu:
Another bug fixed
And according to my test, the figure shown on the [website](https://pypose.org/docs/main/generated/pypose.optim.kernel.Tolerant/) is correct

pypose/optim/kernel.py
```diff
@@ -255,7 +255,7 @@ def forward(self, input: Tensor) -> Tensor:
         '''
         assert torch.all(input >= 0), 'input has to be non-negative'
         part1 = (1 + ((input-self.a) / self.b).exp()).log()
-        part2 = (1 + (-self.a / self.b).exp())
+        part2 = (1 + torch.tensor(-self.a / self.b).exp()).log()
```
wang-chen, line 258: the entire item is a scalar, we can pre-compute it in `__init__`  by using the `math` library, as the overhead of small tensor is big.

pypose/optim/kernel.py
```diff
@@ -245,18 +245,17 @@ def __init__(self, a: float = 1.0, b: float = -1.0) -> None:
...
         Args:
             input (torch.Tensor): the input tensor (non-negative).
         '''
         assert torch.all(input >= 0), 'input has to be non-negative'
-        part1 = (1 + ((input-self.a) / self.b).exp()).log()
-        part2 = (1 + (-self.a / self.b).exp())
-        return self.b * part1 - self.b * part2
+        res = self.b * (1 + ((input-self.a) / self.b).exp()).log()
+        offset = self.b * math.log((1 + math.exp(-self.a / self.b)))
+        return res - offset
```
wang-chen, line 258: How about
```
result = self.b * (1 + ((input-self.a) / self.b).exp()).log()
offset = self.b * math.log((1 + math.exp(-self.a / self.b))) # constant, no grad.
return result - offset
```

ShaoshuSu: You mean replacing `res` with `result`?

wang-chen: yes, this will make the two variables have same name length.
And add the short comments to explain why we use `math` library not `torch` library.

wang-chen (approved):

Issue #148, opened 2022-10-14
Optimizers accepts model output as tuple.


wang-chen:
### 🚀 The feature, motivation and pitch

The current [LM](https://pypose.org/docs/main/generated/pypose.optim.LevenbergMarquardt/#pypose.optim.LevenbergMarquardt) and [GN](https://pypose.org/docs/main/generated/pypose.optim.GaussNewton/#pypose.optim.GaussNewton) optimizers only minimize single-model output as a tensor.
Refactor is to accept both tensor and tuple of tensors so that other constraints can be added.

PR #243, opened 2023-05-19
Feature/multi error

xukuanHIT:
#148


pypose/optim/optimizer.py
```diff
@@ -32,22 +32,34 @@ def __init__(self, model, kernel=None, auto=False):
...
+            return None
+        data_list = [data.reshape(-1) for data in data_list]
+        return torch.cat(data_list)
+
     def residual(self, output, target):
         return output if target is None else output - target
 
+    def residuals(self, output, target):
+        if isinstance(output, (tuple, list)):
+            target = [None for i in range(len(output))] if target is None else target
```
wang-chen, line 59: how about
```python
target = [None] * len(output) if target is None else target
```

pypose/optim/optimizer.py
```diff
@@ -32,22 +32,34 @@ def __init__(self, model, kernel=None, auto=False):
...
+            return tuple([self.residual(output[i], target[i]) for i in range(len(output))])
```
wang-chen, line 60: how about
```python
return tuple([self.residual(out, target[i]) for i, out in enumerate(output)])
```

pypose/optim/optimizer.py
```diff
@@ -32,22 +32,34 @@ def __init__(self, model, kernel=None, auto=False):
...
+    def merge_data_list(self, data_list):
+        if data_list is None:
+            return None
+        data_list = [data.reshape(-1) for data in data_list]
+        return torch.cat(data_list)
+
     def residual(self, output, target):
         return output if target is None else output - target
 
+    def residuals(self, output, target):
```
wang-chen, line 57: how about change args name to `outputs` and `targets`

pypose/optim/optimizer.py
```diff
@@ -32,22 +32,34 @@ def __init__(self, model, kernel=None, auto=False):
...
             return self.model(*input)
         else:
             return self.model(input)
 
+    def merge_data_list(self, data_list):
+        if data_list is None:
+            return None
+        data_list = [data.reshape(-1) for data in data_list]
+        return torch.cat(data_list)
+
```
wang-chen, line 53: it seems that this function is not used, remove it.

pypose/optim/optimizer.py
```diff
@@ -56,8 +68,12 @@ def kernel_forward(self, module, input, output):
...
     def loss(self, input, target):
         output = self.model_forward(input)
-        residual = self.residual(output, target)
-        return self.kernel(residual.square().sum(-1)).sum()
+        residual_list = self.residuals(output, target)
+        if isinstance(self.kernel, nn.ModuleList):
+            residuals = [k(r.square().sum(-1)).sum() for k, r in zip(self.kernel, residual_list)]
+        else:
+            residuals = [self.kernel(r.square().sum(-1)).sum() for r in residual_list]
+        return sum(residuals)
```
wang-chen, line 76: how about change `residual_list` to `residuals`

pypose/optim/optimizer.py
```diff
@@ -175,6 +191,19 @@ def __init__(self, model, solver=None, kernel=None, corrector=None, weight=None,
...
 
+    def normalize_R_weight(self, R, weight=None):
+        weight_diag = None
+        if weight is not None:
+            weight = weight if isinstance(weight, (tuple, list)) else [weight]
+            assert len(R)==len(weight)
+            weight_diag = []
+            for w, r in zip(weight, R):
+                ni = r.numel() / w.shape[-1]
+                weight_diag += [w for i in range(int(ni))]
```
wang-chen, line 202: do you mean
```
weight_diag += [w] * int(ni)
```

pypose/optim/optimizer.py
```diff
@@ -175,6 +191,19 @@ def __init__(self, model, solver=None, kernel=None, corrector=None, weight=None,
             self.corrector = Trivial() if corrector is None else corrector
         self.weight = weight
 
+    def normalize_R_weight(self, R, weight=None):
```
wang-chen, line 194: is this function identical to both GN and LM? if yes, how about moving it to the class of `RobustModel`, so that it only appear once?

wang-chen: I also need a meeting with you to explain the math background of this function.

wang-chen (changes requested):

pypose/optim/optimizer.py
```diff
@@ -32,22 +32,34 @@ def __init__(self, model, kernel=None, auto=False):
...
 
+    def residuals(self, output, target):
+        if isinstance(output, (tuple, list)):
+            target = [None for i in range(len(output))] if target is None else target
+            return tuple([self.residual(output[i], target[i]) for i in range(len(output))])
+        return tuple([self.residual(output, target)])
+
     def kernel_forward(self, module, input, output):
         # eps is to prevent grad of sqrt() from being inf
         assert torch.is_floating_point(output), "model output have to be float type."
```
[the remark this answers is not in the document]
xukuanHIT: OK

pypose/optim/optimizer.py
xukuanHIT: OK

pypose/optim/optimizer.py
xukuanHIT: ok

PR #244, opened 2023-05-23
Add operation for SBT

ZihangFang:


pypose/sparse/sparse_block_tensor.py
```diff
@@ -425,6 +425,30 @@ def storage_post(self, func, types, s_outs=(), p_outs=(), kwargs={}):
         '''
         return s_outs, p_outs
 
+class ProxyAddOp(SBTProxyNoOp):
```
huyaoyu, line 428: Why do we need this `AddOp` and why does it inherit from `NoOp`? `NoOp` Oringially means no operation is needed for the proxy.

huyaoyu (commented):
Please see the comments. Thanks.

pypose/sparse/sparse_block_tensor.py
zitongzhan: The ProxyAddOp is to support operations like `add`, `sub`. Suppose we want `sbt1 + sbt2`, then we need to find the non-zero blocks that either exist in `sbt1` or `sbt2`. In other words, the resulting sbt's proxy is an `logical_or` operation on the proxies of original operants, and use the `__add__` operant provided by COO Hybrid for computation of storage. 

The only difference between NoOp and AddOp is its proxy computation, but AddOp has the exact same storage computation as NoOp, thus AddOp inherit from NoOp.

huyaoyu (approved):
Generally good. 

I might have some suggestions on the naming of things, though. 

Let's merge it for the moment.

PR #249, opened 2023-06-04
Ipddp

wang-chen:
ipddp by @ntu-caokun


wang-chen:
Some points to fix conflicts before further review and merge:
- new functions like `bdot` can be implemented via 'torch.einsum'. The new functions are not used by LieTensor, thus cannot be defined in `basics`, which will cause circular imports error. Instead, define them in `function/linalg.py`
- Don’t import `pypose` in pypose modules, only import pypose in tests and examples. See 14 in [developer guide](https://github.com/pypose/pypose/wiki/Developer-Guideline)
- Follow `dynamics.py` on main branch, better not to change it.
- For testing codes, please follow tests style on the main branch.

ntu-caokun:
auto-tests has passed, plz review @aerogjy

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+        >>> for batch_id in range(n_batch): # use for loop and keep the IPDDP
+        >>>     stage_cost = pp.module.QuadCost(Q[batch_id:batch_id+1], R[batch_id:batch_id+1], S[batch_id:batch_id+1], c[batch_id:batch_id+1])
+        >>>     terminal_cost = pp.module.QuadCost(10./dt*Q[batch_id:batch_id+1,0:1,:,:], R[batch_id:batch_id+1,0:1,:,:], S[batch_id:batch_id+1,0:1,:,:], c[batch_id:batch_id+1,0:1])
+        >>>     lincon = pp.module.LinCon(gx[batch_id:batch_id+1], gu[batch_id:batch_id+1], g[batch_id:batch_id+1])
+        >>>     init_traj_sample = {'state': init_traj['state'][batch_id:batch_id+1], 'input': init_traj['input'][batch_id:batch_id+1]}
+        >>>     solver = IPDDP(sys, stage_cost, terminal_cost, lincon, gx.shape[-2], init_traj_sample)
+        >>>     traj_opt[batch_id] = solver.optimizer()
+
+    '''
+    def __init__(self, sys=None, stage_cost=None, terminal_cost=None, cons=None, n_cons=0, init_traj=None):
```
aerogjy, line 184: The API design should be similar to lqr and mpc, which consists of the system, time horizon and the cost related terms. Here, I think the cons and n_cons could be somehow merged together? Also the init_traj shouldn't be necessary.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+                self.mu = max(self.tol/10.0, min(0.2*self.mu, pow(self.mu, 1.2) ) )
+                self.resetfilter()
+                self.reg, self.bp_failed, self.recovery = 0.0, False, 0
+
+            if iter == self.maxiter - 1:
+                print("max iter", self.maxiter, "reached, not the optimal one!")
+            #-----------------------------------------------
+        return self
+
+    def forward(self, fp_list):
```
aerogjy, line 530: I'm kind of confused about solver and forward. It seems in the example, forward is used. In the test case, solver is used. I suggest we keep the consistency with the lqr and mpc with a forward would be great. Also, try to keep similar arg list of the forward function.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+
+    where :math:`\mathbf{\tau}_t` = :math:`\begin{bmatrix} \mathbf{x}_t \\ \mathbf{u}_t
+    \end{bmatrix}`; :math:`q` and :math:`p` denote the stage and terminal costs, respectively;
+    :math:`\mathbf{c}` is the stage-wise inequality constraints. In addition to the notations
+    defined above, two additional variables :math:`\mathbf{s}` and :math:`\mathbf{y}` are
+    introduced, which denote the dual variable and the slack variable, respectively.
+
+    The IPDDP process can be summarised as iterative backward and forward recursions.
+
+    - The backward recursion.
```
aerogjy, line 48: It seems two scenarios are implemented with one starts from a feasible trajectory and the other from an infeasible trajectory. But only one scenario's doc is provided? That's why there is no updated related to the slack variable y, although it's defined.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+
+        self.constraint_flag = True
+        self.contraction_flag = True
+
+        self.x, self.u = init_traj['state'], init_traj['input']
+        B = self.x.shape[:-2]
+        ns, nc, ncons, self.T = self.x.size(-1), self.u.size(-1), n_cons, self.u.size(-2)
+
+        # algorithm parameter
+        self.mu, self.maxiter, self.tol, self.infeas = 1.0, 50, torch.tensor([1.0e-7], dtype=self.x.dtype, device=self.x.device), False
```
aerogjy, line 199: For self.infeas, didn't see anywhere check the start trajectory and set it to true or false.

pypose/module/constraint.py
```diff
@@ -0,0 +1,197 @@
...
+        Constant term generated by constraint.
+
+        .. math::
+            \mathbf{g} = \mathrm{cons}(\mathbf{x}^*, \mathbf{u}^*, t^*)
+                           - \mathbf{g}_{\mathbf{x}}\mathbf{x}^* - \mathbf{g}_{\mathbf{u}}\mathbf{u}^*
+        '''
+        # Potential performance loss here - self.A and self.B involves jacobian eval
+        return self._ref_c - pp.bmv(self.gx, self._ref_state) - pp.bmv(self.gu, self._ref_input)
+
+class LinCon(Constraint):
```
aerogjy, line 105: Is this constraint inequality constraint? Didn't see anywhere in the doc defines that and whether it's >0, or <0, or >=0, or <=0.

examples/module/ipddp/train_ipddp_invpend.py
```diff
@@ -0,0 +1,163 @@
...
+            # detached version to solve the best traj
+            with torch.no_grad():
+                _fp_best_list[batch_id] = _ipddp_list[batch_id].solver()
+
+        x_pred, u_pred, _ = _ipddp_list[0].forward(_fp_best_list) # call any one class instantiation, but perform batch grad computation
+        print('x_pred solved')
+        sys = InvPend(dt, length=expert['param'])
+        ipddp_list = [None for batch_id in range(n_batch)]
+        fp_list = [None for batch_id in range(n_batch)]
+        for batch_id in range(n_batch):
```
aerogjy, line 112: There should be some better way to handle the batch scenario. I suggest we focus on the non-batched case first.

examples/module/ipddp/train_ipddp_invpend.py
```diff
@@ -0,0 +1,163 @@
...
+    device = 'cuda' if args.cuda else 'cpu'
+
+    ns, nc = args.n_state, args.n_input
+    T = args.T
+    expert_seed = 26
+    state = torch.tensor([[-2.,0.],
+                          [-1., 0.],
+                          [-2.5, 1.]],
+                          device=device)
+    state = torch.tensor([[-2.,0.],
```
aerogjy, line 43: Why define state twice?

tests/module/test_ipddp.py
```diff
@@ -0,0 +1,71 @@
...
+
+    for batch_id in range(n_batch): # use for loop and keep the IPDDP
+        # used class is for batched version, inside loop use batch_size = 1
+        stage_cost = pp.module.QuadCost(Q[batch_id:batch_id+1], R[batch_id:batch_id+1], S[batch_id:batch_id+1], c[batch_id:batch_id+1])
+        terminal_cost = pp.module.QuadCost(10./dt*Q[batch_id:batch_id+1,0:1,:,:], R[batch_id:batch_id+1,0:1,:,:], S[batch_id:batch_id+1,0:1,:,:], c[batch_id:batch_id+1,0:1])
+        lincon = pp.module.LinCon(gx[batch_id:batch_id+1], gu[batch_id:batch_id+1], g[batch_id:batch_id+1])
+        init_traj_sample = {'state': init_traj['state'][batch_id:batch_id+1],
+                            'input': init_traj['input'][batch_id:batch_id+1]}
+        ipddp = IPDDP(sys, stage_cost, terminal_cost, lincon, gx.shape[-2], init_traj_sample)
+        traj_opt[batch_id] = ipddp.solver()
```
aerogjy, line 71: Here, used solver, but in the training example, used forward.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+
+        # quantities in forward pass
+        self.c = torch.zeros(B + (self.T, ncons), dtype=self.x.dtype, device=self.x.device)
+        self.s = 0.1 * torch.ones(B + (self.T, ncons), dtype=self.x.dtype, device=self.x.device) # dual variable for inequality constraint c
+        self.y = 0.01 * torch.ones(B + (self.T, ncons), dtype=self.x.dtype, device=self.x.device) # slack variable for infeasible case
+
+        # -------- derivatives --------------------------
+        # terms related with system dynamics
+        self.fx = torch.zeros(B + (self.T, ns, ns), dtype=self.x.dtype, device=self.x.device)
```
aerogjy, line 208: Line 208 to 237 are a bit overwhelm. I think there should be a better way without defining all of the variables at the beginning. Maybe @wang-chen can have some suggestion.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+        self.ky = torch.zeros(B + (self.T, ncons), dtype=self.x.dtype, device=self.x.device)
+        self.Ky = torch.zeros(B + (self.T, ncons, ns), dtype=self.x.dtype, device=self.x.device)
+        self.ks = torch.zeros(B + (self.T, ncons), dtype=self.x.dtype, device=self.x.device)
+        self.Ks = torch.zeros(B + (self.T, ncons, ns), dtype=self.x.dtype, device=self.x.device)
+        self.ku = torch.zeros(B + (self.T, nc), dtype=self.x.dtype, device=self.x.device)
+        self.Ku = torch.zeros(B + (self.T, nc, ns), dtype=self.x.dtype, device=self.x.device)
+        self.opterr, self.reg, self.bp_failed, self.recovery = 0., 0., False, 0
+
+    def getDerivatives(self):
+        self.p_fn.set_refpoint(self.x[...,-1,:], self.u[...,-1,:])
```
aerogjy, line 241: Same thing here. Can we just directly call the derivative when you are using them? Since you've already defined separate class for the cost and constraints.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,573 @@
...
+            ns, nc = self.Qx_terminal.size(-1), self.F.size(-1) - self.Qx_terminal.size(-1)
+            self.Ku = torch.zeros(B + (self.T, nc, ns), dtype=self.p.dtype, device=self.p.device)
+            self.ku = torch.zeros(B + (self.T, nc), dtype=self.p.dtype, device=self.p.device)
+
+        ns = self.x.shape[-1]
+        if not lastIterFlag:
+            c_err, r_err, qu_err = torch.zeros(B, dtype=self.x.dtype, device=self.x.device), torch.zeros(B, dtype=self.x.dtype, device=self.x.device), torch.zeros(B, dtype=self.x.dtype, device=self.x.device)
+
+            # set regularization parameter
+            if (self.fp_failed or self.bp_failed):
```
aerogjy, line 303: Not recommend. @wang-chen how to avoid those long parameter discussion?

aerogjy (changes requested):
I can successfully run all the examples and tests. Several major issues: 1. The API design should be consistent with lqr or mpc. 2. Infeasible start traj is missing in the docs. 3. The constraints type definition is a bit unclear. 4. Not sure the current organization of cost and constraints in the modules are proper @wang-chen 5. Some of the code in the core ipddp class is overwhelming, need to clean up, and try to avoid long variable discussion. (See the detailed comments inline the code.)

tests/module/test_ipddp.py
ntu-caokun: In the test, it only cares about solving the traj, and does not create a computational graph for learning; in the training example, it first calls solver() to solve the trajectory, then use forward to create the computational graph for autodiff.

pypose/module/ipddp.py
ntu-caokun: In the test, it only cares about solving the traj, and does not create a computational graph for learning; in the training example, it first calls solver() to solve the trajectory, then use forward to create the computational graph for autodiff.

pypose/module/constraint.py
ntu-caokun: Added to the docs in constraint.py

examples/module/ipddp/train_ipddp_invpend.py
ntu-caokun: fixed it.

examples/module/ipddp/train_ipddp_invpend.py
ntu-caokun: Yes, currently the solver() can only solve a single trajectory, but the autodiff part using forward() supports batched case; so I leave the interface of the batched case here. 
If the users are interested in the non-batched case, they can just set the state with n_batch = 1.

pypose/module/ipddp.py
ntu-caokun: that will make the backwardpasscompact() overwhelm. Notice that in LQR, the cost-related terms are passed in instead of being computed inside the loop, and there are no second-order dynamics terms and constraint terms, which makes the LQR code much more elegant.

pypose/module/ipddp.py
ntu-caokun: added

pypose/module/ipddp.py
ntu-caokun: merged n_cons;
time_horizon maybe not be necessary, if time-varying LQR is considered, then Q matrix should be of shape [ns,ns,T], T is included in Q;
in my case, init_traj seems to be necessary, as I need it to extract ns, nc, and initialize many matrices; while lqr initializes these matrices in lqr_backward, which is not suitable for our case where backward will be called many times.

pypose/module/ipddp.py
ntu-caokun: fixed this.

ntu-caokun:
fixed issues 2, 3;
issue 1, 4, 5 needs discussion

ntu-caokun:
Please use the [da9aaf6](https://github.com/pypose/pypose/pull/249/commits/da9aaf66875d37d615a5056efb003b0da18315a9) to review, as the latest merge cannot pass the tests;
In new commits: as per discussion, I have unified the init function interface with MPC, while still adding the B argument, for batch size;
tidy up the code.

examples/module/ipddp/train_ipddp_invpend.py
```diff
@@ -0,0 +1,149 @@
+import os
+import time
+import torch
+import shutil
+import argparse
+import pypose as pp
+import pickle as pkl
+import torch.optim as optim
+from pypose.module.ipddp import IPDDP
+from tests.module.test_ipddp import InvPend
```
aerogjy, line 10: Could we make a InvPend as a dynamic example in the pypose/examples, rather than in the test_ipddp? That will make our code more general and modular to be used by other users.

pypose/module/ipddp.py
```diff
@@ -0,0 +1,576 @@
...
+
+        .. math::
+            \mathbf{cost} \left( \mathbf{\tau}_t \right) = \sum\limits_{t=0}^{T-1}q(\mathbf{\tau}_t) + p(\mathbf{\tau}_T)
+
+    Note:
+        Some additional tricks were used in the implementation, i.e., egularization in the backwardpass,
+        filter, line search in the forwardpass.
+
+    From the learning perspective, this can be interpreted as a module with unknown parameters in
+    :math:`\begin{Bmatrix} \mathbf{f}, \mathbf{c}, q, p \end{Bmatrix}`,
```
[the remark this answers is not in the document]
ntu-caokun: q, p here denote the stage and terminal costs, respectively.

pypose/module/constraint.py
```diff
@@ -0,0 +1,195 @@
...
+        '''
+        return self._cdim
+
+    @property
+    def gx(self):
+        r'''
+        Linear/linearized constraint term on state
+
+        .. math::
+            \mathbf{g}_{\mathbf_{x}} = \left. \frac{\partial \mathbf{g}}{\partial \mathbf{x}} \right|_{\chi^*}
```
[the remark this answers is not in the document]
ntu-caokun: merged solver and forward;
previously solver is done for each sample, and then collect them together in batch and do forward to get computational graph;

now each sample use their own instance, i.e., do their solve and computational graph separately.

PR #250, opened 2023-06-05
Support controller parameters tuning pipeline

xiangfuli:
The tuner is built based on the sensitivity propagation optimization, the related paper can be found in [here](https://arxiv.org/abs/2209.10021).

To use the tuner, four components must be defined by the user:
1. The dynamic model
2. The linear/non-linear controller
3. The desired states for the model
4. The function to get the error between desired states and current states


examples/module/controller/SE3_controller.py
```diff
@@ -0,0 +1,60 @@
+import torch
+from pypose.module.controller import Controller
+from examples.module.controller_parameters_tuner.commons \
+  import hat, vee, quaternion_2_rotation_matrix
+
+
+class SE3Controller(Controller):
```
aerogjy, line 7: Suggest using geometric control naming, and move to the module rather than staying in example

examples/module/controller/SE3_controller.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # extract parameters
+        kp, kv, kori, kw = parameters
+
+        pose = torch.t(torch.atleast_2d(pose))
+        angular_vel = torch.t(torch.atleast_2d(angular_vel))
+        desired_angular_vel = torch.t(torch.atleast_2d(desired_angular_vel))
+        desired_angular_acc = torch.t(torch.atleast_2d(desired_angular_acc))
+
+        Rwb = quaternion_2_rotation_matrix(pose)
```
aerogjy, line 34: Let's delete all these conversion and use LieTensor for all kinds of operation.

examples/module/controller/SE3_controller.py
```diff
@@ -0,0 +1,60 @@
...
+
+        M = - kori * err_pose - kw * err_angular_vel \
+          + torch.cross(angular_vel, torch.mm(self.J, angular_vel))
+        temp_M = torch.mm(hat(angular_vel),
+                           torch.mm(Rwb.T, torch.mm(desired_pose, desired_angular_vel))) \
+                            - torch.mm(Rwb.T, torch.mm(desired_pose, desired_angular_acc))
+        M = M - torch.mm(self.J, temp_M)
+
+        return torch.concat(
+          [torch.max(torch.tensor([0.], device=device), f[0]), M[0], M[1], M[2]]
```
aerogjy, line 59: Find another nicer way of this return.

examples/module/controller_parameters_tuner/commons.py
```diff
@@ -0,0 +1,88 @@
+import torch
+
```
aerogjy, line 2: Like I said before, we can use Lietensor's built in functionality for these common operation.

examples/module/controller_parameters_tuner/dubincar_controller_tuner.py
```diff
@@ -0,0 +1,153 @@
...
+          ref_position[1],
+          ref_pose,
+          torch.norm(ref_velocity[0]),
+          ref_angular_vel
+       ])
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Cartpole Example')
```
aerogjy, line 91: Not cartpole

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,219 @@
...
+    parser.set_defaults(show=False)
+    args = parser.parse_args(); print(args)
+    os.makedirs(os.path.join(args.save), exist_ok=True)
+
+    g = 9.81
+
+    # program parameters
+    time_interval = 0.05
+    learning_rate = 0.01
+    initial_state = torch.tensor([
```
aerogjy, line 140: These are so ugly....

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,219 @@
...
+        torch.tensor([0.]),
+        torch.tensor([0.]),
+        torch.tensor([1.])]
+      ).to(device=args.device).double()
+    multicopter = MultiCopter(time_interval,
+                               0.6,
+                               torch.tensor(g),
+                               torch.tensor([
+                                  [0.0829, 0., 0.],
+                                  [0, 0.0845, 0],
```
aerogjy, line 180: Make a new variable J for the moment of inertia

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,219 @@
...
+                                  [0, 0, 0.1377]
+                                ], device=args.device),
+                                e3)
+
+    # start to tune the controller parameters
+    max_tuning_iterations = 30
```
aerogjy, line 186: The paper uses a more detailed termination condition, such as loss, instead of hard coded iterations

examples/module/dynamics/dubincar.py
```diff
@@ -0,0 +1,47 @@
...
+        k2 = self.derivative(self.euler_update(state, k1, self.tau / 2), input)
+        k3 = self.derivative(self.euler_update(state, k2, self.tau / 2), input)
+        k4 = self.derivative(self.euler_update(state, k3, self.tau), input)
+
+        return state + (k1 + 2 * k2 + 2 * k3 + k4) / 6 * self.tau
+    
+    def observation(self, state, input, t=None):
+        return state
+
+    def derivative(self, state, input):
```
aerogjy, line 21: We can call this xdot rather than derivative....

pypose/optim/controller_parameters_tuner.py
```diff
@@ -0,0 +1,101 @@
...
+
+            dfduk_func = lambda inputs: dynamic_system.state_transition(state = system_state, input = inputs, t = tau)
+            dfduk_tensor = torch.squeeze(jacobian(dfduk_func, controller_input))
+
+            states.append(system_new_state)
+            system_state = system_new_state
+
+            last_gradient = dxdparam_gradients[-1]
+
+            dxdparam_gradients.append(
```
aerogjy, line 83: Add the partial derivative with respect of u_k to accommodate for the general Loss function. such as the one with control penalty term.

pypose/optim/controller_parameters_tuner.py
```diff
@@ -0,0 +1,101 @@
...
+    these components are differentiable, it is possbile to find the most suitable
+    parameters by constructing the loss function $\mathbf{L}$ and computing the gradient of
+    $\mathbf{L}$ with respect to controller parameters.
+
+    Note:
+        For the parameter initial_state and parameters, please use row/column vector to
+        represent these variables at the same time.
+
+    .. math::
+        \begin{align*}
```
aerogjy, line 21: Make the documentation more clear, with all the core algorithms and equations, then add the important citation. Refer to LQR and all other docs.

pypose/optim/controller_parameters_tuner.py
```diff
@@ -0,0 +1,101 @@
+import torch
+from torch import nn
+from torch.autograd.functional import jacobian
+
+
+class ControllerParametersTuner(nn.Module):
```
aerogjy, line 6: Make it general and move to module

aerogjy (changes requested):
First round of review, please see the detailed review in the code. Several major issues: 1. Use LieTensor rather than re-write all the conversion by yourself. 2.Compute the partial derivative for u_k to accommodate the general loss L instead of only for x_k for sensitivity propogation. 3. Re-organize the modules , put the controller and tuner to modules rather than examples or optim. 4. Documentation should have all the core algorithms and equations with properly cited papers. 5. Some coding style really needs improved. Let's fix these before further review.

xiangfuli:
According to the comments, several revises have been made:

* Use the `skew2vec`, `vec2skew` function provided in the pypose. But still keeping the `quaternion_2_rotation_matrix` and `rotation_matrix_2_quaternion` function since pypose does not provide such function.
* Add the controller inputs $u_k$ as the penalty in the loss function
* Have moved the controllers: `dubincar controller` and `geometric controller` to the directory `pypose.module.controllers`, also have moved `ControllerParametersTuner` into the directory `pypose.module`
* Parts of the codes have been polished.

ToDo:
* Working on the documentation.

xiangfuli:
Documentation has been updated on Wednesday.

pypose/lietensor/basics.py
```diff
@@ -41,6 +41,40 @@ def vec2skew(input:torch.Tensor) -> torch.Tensor:
...
+    Note:
+        The last 2 dimension of the input tensor has to be (3, 3).
+
+    Example:
+        >>> pp.skew2vec(torch.randn(1, 3, 3))
+        tensor([[-0.2929, -1.2761, 2.2059]])
+    """
+    v = input.tensor() if hasattr(input, 'ltype') else input
+    assert v.shape[-2:] == (3, 3), "Last 2 dim should be (3, 3)"
+    # assert torch.equal(v.permute(0, 2, 1), -v), "Each matrix must be a skew matrix"
```
xiangfuli, line 72: This line is commented because with using cuda devices, when calculating the matrix C with torch.mm(A, B.T) - torch.mm(B, A.T), the result would not even meet with C == -C.T. But when use CPU to perform such operations, the result would meet with C == -C.T.

xiangfuli:
@wang-chen Hello Chen, In this PR, I am trying to implement the function `skew2vec`(also mentioned in the  #255) in the file [pypose/lietensor/basics.py](https://github.com/pypose/pypose/blob/37f3542f4ff5770163c52357492ced48ba3e219e/pypose/lietensor/basics.py#L44C12-L44C12). When I was trying to implement the corresponding unit tests, there is no `test_basics.py` under the test folder. 

Should I also create the unit test for the `skew2vec` function? Thanks.

wang-chen:
Yes, you may do that. Thank you!

xiangfuli:
ToDo:

* Add simple result plots in the new dynamic system examples.
* Add result plots in the tuner examples.

pypose/module/controller_parameters_tuner.py
```diff
@@ -0,0 +1,194 @@
...
+    .. math::
+        \begin{align*}
+            \nabla L_{\theta} &= \sum_{i=1}^N \frac{\partial L}{\partial x_i} \frac{\partial x_i}{\partial \theta} + \sum_{i=0}^{N-1} \frac{\partial L}{\partial u_i} \frac{\partial u_i}{\partial \theta} \\
+            \frac{\partial x_{i+1}}{\partial \theta} &= (\frac{\partial x_{i+1}}{\partial x_i} + \frac{\partial x_{i+1}}{\partial u_i}\frac{\partial u_i}{\partial x_i})
+            \frac{\partial x_i}{\partial \theta}+\frac{\partial x_{i+1}}{\partial u_i}\frac{\partial u_i}{\partial \theta} \\
+            \frac{\partial u_i}{\partial \theta} &= \frac{\partial u_i}{\partial x_i} \frac{\partial x_i}{\partial \theta} + \frac{\partial u_i}{\partial \theta}
+        \end{align*} \tag{4}
+
+    To make sure the controller parameters can stay in the reasonable and safe sets, after updating the parameters, it's essential to project the new parameters into
+    the feasiable set. And the parameters can be updated using the following equation:
```
aerogjy, line 61: The following update rule doesn't tell you are projecting the new parameters into the feasible set. It's just the normal gradient descent.

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,268 @@
...
+
+    time = torch.arange(0, len(points[0]) - 1 + time_interval, time_interval).numpy()
+    f, ax = plt.subplots(nrows=6, ncols=1, sharex=True)
+    subPlot(ax[0], time, get_sub_states(original_system_states, 0), get_sub_ref_states(ref_states, 0, 0), ylabel='X position')
+    subPlot(ax[1], time, get_sub_states(original_system_states, 1), get_sub_ref_states(ref_states, 0, 1), ylabel='Y position')
+    subPlot(ax[2], time, get_sub_states(original_system_states, 2), get_sub_ref_states(ref_states, 0, 2), ylabel='Z position')
+    subPlot(ax[3], time, get_sub_states(new_system_states, 0), get_sub_ref_states(ref_states, 0, 0), ylabel='X after tuning')
+    subPlot(ax[4], time, get_sub_states(new_system_states, 1), get_sub_ref_states(ref_states, 0, 1), ylabel='Y after tuning')
+    subPlot(ax[5], time, get_sub_states(new_system_states, 2), get_sub_ref_states(ref_states, 0, 2), ylabel='Z after tuning')
+    figure = os.path.join(args.save + 'multicoptor_controller_tuner.png')
```
aerogjy, line 263: The results doesn't seem to track the reference nicely.

aerogjy (changes requested):
I can successfully run all the examples and the test files. Several major issues: 1. bad file organization in the pypose/module. 2. The controller is too problem specific. We are expecting the general PID controller, etc so that people can pass through different problem (dynamics). 3. Many of the API design are not good. Please think about it from the user perspective. 4. Some of the coding style is really not ideal. Please refer to our [developer guideline](https://github.com/pypose/pypose/wiki/Developer-Guideline) SERIOUSLY and CAREFULLY.

xiangfuli:
Setting the pid parameters as the PID controller's internal variables would cause some inconvenience when calculating the gradient inside the parameter tuner. 

We use the `torch.autograd.functional.jacobian` function provided by the PyTorch to get the gradient. When calculating the gradient of controller output w.r.t. the controller parameters, we have to let the controller parameters used as the function input parameters like this: `dhdparam_func = lambda params: controller.get_control(parameters = params,
                                                                  state = system_state, ....)`. Then use the `jacobin` function to get the gradient. `jacobian(dhdparam_func, controller_parameters)`.

 If we set pid parameters as PID's internal properties, we cannot set the `controller_parameters` as the input like this `jacobian(dhdparam_func, controller_parameters)` which causes us cannot use the `jacobian` function anymore.

There is one way to get the gradients properly but it looks much more complex than using the `jacobian` function and needs to modify the existing runner codes a lot. 

Here is the code example (just for showing the new way to get the gradient if set the controller parameters as PID's internal parameters):

```
#!/usr/bin/python3
import torch
from torch.autograd.functional import jacobian

class PID:
    def __init__(self, params) -> None:
        self.params = params
        self.params.requires_grad=True

    def get_control(self, state):
        return state + self.params * 2

parameters = torch.ones(4, requires_grad=True)
pid = PID(parameters)

print("Gradient compution using parameters tensors inside the pid controller: ")
dhdx_func = lambda state: pid.get_control(state)
dhdparam_func = lambda params: pid.get_control(state = torch.ones(4) * 3.5)
print(jacobian(dhdparam_func, parameters))

print("Gradient compution using undetached parameters tensors: ")
new_state = pid.get_control(state = torch.ones(4) * 3.5)
new_state[0].backward(retain_graph=True)
print(pid.params.grad)
new_state[1].backward(retain_graph=True)
print(pid.params.grad)


print("Gradient compution using detached parameters tensors: ")
new_param = torch.clone(parameters).detach()
pid.params = new_param
pid.params.requires_grad=True
new_state0 = pid.get_control(state = torch.ones(4) * 3.5)
new_state0[0].backward(retain_graph=True)
print(pid.params.grad)
new_param = torch.clone(parameters).detach()
pid.params = new_param
pid.params.requires_grad=True
new_state1 = pid.get_control(state = torch.ones(4) * 3.5)
new_state1[1].backward(retain_graph=True)
print(pid.params.grad)
```

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,268 @@
...
+def compute_loss(dynamic_system, controller, controller_parameters, penalty_coefficient,
+                 initial_state, ref_states, dt):
+    loss = 0
+    system_state = torch.clone(initial_state)
+    for index, ref_state in enumerate(ref_states):
+      ref_position, ref_velocity, ref_acceleration, \
+          ref_pose, ref_angular_vel, ref_angular_acc = ref_state
+      controller_input = \
+          controller.get_control(controller_parameters, system_state, ref_state, None)
+      system_new_state = dynamic_system.state_transition(system_state,
```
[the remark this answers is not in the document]
xiangfuli: I was trying to make the tuner looks concise. Right now, I've deleted the loss_computation function and returned the original loss and new loss using new controller parameter in the tuner.

examples/module/controller_parameters_tuner/multicoptor_controller_tuner.py
```diff
@@ -0,0 +1,268 @@
...
+    return loss / len(ref_states)
+
+
+def func_to_get_state_error(state, ref_state):
+    ref_position, ref_velocity, ref_acceleration, \
+      ref_pose, ref_angular_vel, ref_angular_acc = ref_state
+
+    ref_pose_SO3 = pp.from_matrix(ref_pose, ltype=pp.SO3_type).tensor()
+
+    return state - torch.concat(
```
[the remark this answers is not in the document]
xiangfuli: The vector of ref_state may not have the same dimension as the state vector. That's the reason why we may need this error computation function and pass it to the tuner.

Issue #258, opened 2023-07-14
LieTensor becomes torch.Tensor in DataParrallel


zeroAska:
### 🐛 Describe the bug

I'm trying to use run `DataParrallel` for mulit-gpu training with `pypose's LieTensor`. However, I noticed that when I sent a `LieTensor` pose object to the data parrallel model, the pose becomes a `torch.Tensor`:

Example:
```                                                                                                                                      
import torch                                                                                                                              
import torch.nn                                                                                                                           
import pypose as pp                                                                                                                       
class DataParallelModel(nn.Module):                                                                                                       
    def __init__(self):                                                                                                                   
        super().__init__()                                                                                                                
                                                                                                                                          
    def forward(self, x, T):          
        print("type of T is ",type(T))                                                                                                    
        assert(isinstance(T, pp.LieTensor))                                                                                               
        return T @ x                                                                                                                      
                                                                                                                                          
def test_parallel():                                                                                                                      
    a = pp.randn_SO3(4)                                                                                                                   
    a.requires_grad = True                                                                                                                
                                                                                                                                          
    net = DataParallelModel()                                                                                                             
    if torch.cuda.device_count() > 1:                                                                                                     
        print("Let's use", torch.cuda.device_count(), "GPUs!")                                                                            
        net = nn.DataParallel(net).cuda()                                                                                                 
        net(torch.rand(4), a)                                                                                                             
                                                                                                                                          
test_parallel()
```


Output:
```
Let's use 4 GPUs!
type of T is  <class 'torch.Tensor'>
type of T is  <class 'torch.Tensor'>
type of T is  <class 'torch.Tensor'>
type of T is  <class 'torch.Tensor'>
....
Original Traceback (most recent call last):
  File ".../python3.7/site-packages/torch/nn/parallel/parallel_apply.py", line 64, in _wo
rker
    output = module(*input, **kwargs)
  File ".../python3.7/site-packages/torch/nn/modules/module.py", line 1194, in _call_impl
    return forward_call(*input, **kwargs)
  File "test_backprop.py", line 65, in forward
    assert(isinstance(T, pp.LieTensor))
AssertionError  
```


### Versions

```
% python collect_env.py 
Collecting environment information...
PyTorch version: 1.13.1+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

GCC version: (GCC) 7.3.1 
Clang version: Could not collect
CMake version: Could not collect
[...]

PR #262, opened 2023-07-16
LieTensor workable in func.jacrev

zitongzhan:
initial solution for letting lietensor workable in jacrev


zitongzhan:
@hxu296 Could you please take a look at the PR and maybe limit code redundancy in the context manager? So far this is only a workable solution but still have duplicate code segments. Perhaps we could write a general wrapper/decorator that is workable on any torch functions like this type.

pypose/lietensor/lietensor.py
```diff
@@ -1220,3 +1220,32 @@ def __deepcopy__(self, memo):
...
+    def _wrap_tensor_for_grad(cls, tensor, level):
+        ltype = tensor.ltype if isinstance(tensor, LieTensor) else None
+        res = WrappableLT.tmp_w.__call__(tensor, level)
+        if ltype is not None:
+            res = torch.Tensor.as_subclass(res, LieTensor)
+            res.ltype = ltype
+        return res
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        torch.autograd.forward_ad.make_dual = self.tmp_md
```
hxu296, line 1251: I think we also need to restore the original `torch._functorch.eager_transforms._wrap_tensor_for_grad` here.

Also I think we can use the more modern [contextlib](https://docs.python.org/3/library/contextlib.html) to write the context manager. I did not think of a direct benefit of using contextlib, but it is a more popular option in PyTorch's source code.

pypose/lietensor/lietensor.py
```diff
@@ -1220,3 +1220,32 @@ def __deepcopy__(self, memo):
...
+        torch.autograd.forward_ad.make_dual = self.make_dual_wrapper
+        torch._functorch.eager_transforms._wrap_tensor_for_grad = self._wrap_tensor_for_grad
+    
+    @classmethod
+    def make_dual_wrapper(cls, tensor, tangent, *args, level=None):
+        ltype = tensor.ltype if isinstance(tensor, LieTensor) else None
+        res = WrappableLT.tmp_md.__call__(tensor, tangent, *args, level=level)
+        if ltype is not None:
+            res = torch.Tensor.as_subclass(res, LieTensor)
+            res.ltype = ltype
```
hxu296, line 1238: nit: maybe directly using LieTensor's constructor is cleaner

tests/optim/test_jacobian.py
```diff
@@ -117,6 +119,26 @@ def forward(self, inputs):
...
+            jac_func = jacfwd(func)
+            jac = jac_func(pose, points)
+            raise AssertionError('should not reach here')
+        except RuntimeError as e:
+            assert 'shapes cannot be multiplied' in str(e)
+
+        with WrappableLT():
+            jac_func = jacrev(func)
+            jac = jac_func(pose, points)
+            assert not pp.hasnan(jac)
```
hxu296, line 139: We can assert the jac with LieTensor input is all close to the jac with only Torch.Tensor input if inputs have the same data.

tests/optim/test_jacobian.py
```diff
@@ -117,6 +119,26 @@ def forward(self, inputs):
...
+
+        try: # the torch behavior since ver 2.0.0
+            jac_func = jacfwd(func)
+            jac = jac_func(pose, points)
+            raise AssertionError('should not reach here')
+        except RuntimeError as e:
+            assert 'shapes cannot be multiplied' in str(e)
+
+        with WrappableLT():
+            jac_func = jacrev(func)
```
hxu296, line 137: Is there a reason that on Line 130 we used `jacfwd`, but on Line 137 here we used `jacrev`?

pypose/lietensor/operation.py
```diff
@@ -471,14 +472,20 @@ def backward(ctx, grad_output):
...
-        ctx.save_for_backward(X, out)
         return out
 
+    @staticmethod
+    def setup_context(ctx, inputs, output):
+        X, p = inputs
+        out = output
+        ctx.save_for_backward(X, out)
+        return
+
```
hxu296, line 488: Seems like this is a nice bit of cleanup that is tangent to the context manager we are adding. I am not very familiar with this part of the code base, but would adding a link to PyTorch's documentation about [extending torch.autograd](https://pytorch.org/docs/stable/notes/extending.html) be helpful here? It took me a while to understand why we can safely remove the `ctx` from `forward` by adding the `setup_context` here, and this documentation helped me personally.

hxu296 (commented):
Swapping the torch ops worked, nice! I added some comment to point out some places that look weird to me, but generally this is a great step forward!

hxu296:
I will go ahead and address some of my own comment. After that, we can meet to discuss how to make it more general.

pypose/lietensor/operation.py
zitongzhan: It's just a new requirement from torch 2.0

tests/optim/test_jacobian.py
zitongzhan: jacrev is the only thing we can support for now. jacfwd needs us to implement jvp in SE3_Act. jvp is also briefly discussed in https://pytorch.org/docs/stable/notes/extending.html

pypose/lietensor/lietensor.py
zitongzhan: as_subclass is the official recommended way, if LieTensor constructor did not call `as_subclass `

tests/optim/test_jacobian.py
hxu296: Thanks for the link!

tests/optim/test_jacobian.py
hxu296: Did not find an easy way to do that. It is hard to reproduce the exact computation graph as a LieTensor goes through.

hxu296:
`wrappable_lt` is potentially useful to solve problems like #258

hxu296 (changes requested):
Look good! I modified the retain_ltype context manager to restore functions-to-wrap and added tests to assert functions are indeed restored. Next step is to make the retain_ltype context manager thread safe. The save, swap, and restore critical regions need to be protected with a semaphore.

hxu296 (approved):

pypose/func/jac.py
```diff
@@ -0,0 +1,58 @@
+import torch
+from typing import Callable, Union, Tuple, Optional
+from pypose.lietensor.lietensor import retain_ltype
```
wang-chen, line 3: Don't import pypose in pypose modules, only import pypose in examples and tests.

pypose/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .lietensor import Sim3_type, sim3_type, RxSO3_type, rxso3_type
 from .lietensor import tensor, translation, rotation, scale, matrix, euler
 from .lietensor import mat2SO3, mat2SE3, mat2Sim3, mat2RxSO3, from_matrix, matrix, euler2SO3, vec2skew
+from .func import *
```
wang-chen: suggest
```
from .lietensor.lietensor import retain_ltype
from . import func
```

pypose/func/jac.py
wang-chen: suggest:
```
from .. import retain_ltype
```

pypose/lietensor/lietensor.py
```diff
@@ -1,7 +1,8 @@
 import torch
 from torch import nn
+from contextlib import contextmanager
 from .basics import vec2skew
-import collections, numbers, warnings
+import collections, numbers, warnings, importlib
 from .operation import broadcast_inputs
```
wang-chen: use pyramid style, refer to point 12 in [guideline](https://github.com/pypose/pypose/wiki/Developer-Guideline)

tests/optim/test_jacobian.py
```diff
@@ -117,9 +122,69 @@ def forward(self, inputs):
...
+                    assert_fn_equal(func1, func2)
+
+        # save functions to be checked
+        TO_BE_CHECKED = {
+            torch.autograd.forward_ad.make_dual,
+            torch._functorch.eager_transforms._wrap_tensor_for_grad,
+        }
+
+        with check_fn_equal(TO_BE_CHECKED):
+            with retain_ltype():
```
wang-chen: suggest using `pp.retain_ltype()`

wang-chen (approved):

PR #263, opened 2023-07-18
Update lqr backward, use cholesky method to solve Ax = B

QH-Li-1999:
Accelerate around 20% when dimensions higher than 10.


pypose/module/lqr.py
```diff
@@ -345,10 +346,11 @@ def lqr_backward(self, x_init, dt, u_traj=None, u_lower=None, u_upper=None, du=N
             Qxx, Qxu = Qt[..., :ns, :ns], Qt[..., :ns, ns:]
             Qux, Quu = Qt[..., ns:, :ns], Qt[..., ns:, ns:]
             qx, qu = qt[..., :ns], qt[..., ns:]
-            Quu_inv = torch.linalg.pinv(Quu)
 
-            K[...,t,:,:] = Kt = - Quu_inv @ Qux
-            k[...,t,:] = kt = - bmv(Quu_inv, qu)
+            solver = ppos.Cholesky()
```
wang-chen, line 350: define solver in `__init__()`
then use it here will be better

wang-chen (changes requested):
please resolve the conflicts.

wang-chen (approved):

PR #272, opened 2023-08-09
fix bugs in test_spline.py

XiangyuChen0212:


wang-chen:
I'll suggest do something like this: 
```python
def assert_lietensor_close(actual, expected, *args, **kwargs):
    diff = (actual.Inv() @ expected).Log().tensor()
    zero = torch.zeros_like(diff)
    torch.testing.assert_close(diff, zero, *args, **kwargs)
```
Then
```python
data = pp.randn_SE3(2, device=device)
poses = pp.bspline(data, 0.10, True)
assert_lietensor_close(poses[...,[0,-1],:], data[...,[0,-1],:], atol=1e-4, rtol=1e-4)
```
Also better to put the assertation as a pypose function and available as `pypose.testing.assert_lietensor_close`.
Could you create a folder `testing` under `pypose` and add the function and doc?

wang-chen (approved):

PR #273, opened 2023-08-24
LQR parallel Jacobian for time invariant system

QH-Li-1999:
Update the Jacobian computation section for LQR to enable parallel calculation of Jacobians at various time stamps for time-invariant systems.


wang-chen:
Hi, this is a better implementation for Jacobian computation of dynamics systems. I'll suggest replacing the one inside the dynamics system, not only implementing it inside LQR, so that all other classes that use dynamics systems will benefit from it.

PR #279, opened 2023-09-07
try to fix the optimizer bug

xukuanHIT:


xukuanHIT:
There is a risk: when computing the re-projection error using the [provided function](https://github.com/pypose/pypose/blob/64a72bf5df6654e15779c56b3c0e916ff21899cd/pypose/function/geometry.py#L114C5-L114C14), the last dimension would be squeezed if setting "reduction" to "sum" or "norm".

wang-chen (approved):

PR #281, opened 2023-09-20
Fix internal error when both reset and prop_cov are set to False in IMUPreintegrator

luyr:
When initializing IMUPreintegrator, if we set both parameter **prop_cov** and **reset** to False, then an internal error **UnboundLocalError: local variable 'Rij' referenced before assignment** will occur when doing preintegration. 
I add a parameter check in the initialization of IMUPreintegrator, to avoid this internal error.


pypose/module/imu_preintegrator.py
```diff
@@ -98,12 +98,14 @@ def __init__(self, pos = torch.zeros(3),
                        reset = False):
         super().__init__()
         self.reset, self.prop_cov = reset, prop_cov
+        if not self.reset and not self.prop_cov:
+            raise RuntimeError('In IMUPreintegrator, prop_cov must be set to True if reset = False.')
```
wang-chen, line 102: @haleqiu Can you confirm this msg of runtime error is correct?

wang-chen (approved):

PR #293, opened 2023-10-21
distributed matrix multiplication

HongLouyemeng:
I'm not sure if I should use the function directly, or subclass the **torch function**.I'll update the gpu version in the next few days


HongLouyemeng:
Code needs to be cleaned up, I submit functional modules

HongLouyemeng:
tslu can use block-dtensor to reduce code

PR #301, opened 2023-10-27
Update README.md

Aryan4884:
Added comments to describe each part of the code, including the custom neural network, device setup, optimizer, and optimization options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,21 +87,25 @@ pytest
 ```python
 >>> import torch, pypose as pp
 
->>> # A random so(3) LieTensor
+>>> # Create a random so(3) LieTensor with size 2x3x3 and enable gradient tracking
 >>> r = pp.randn_so3(2, requires_grad=True)
     so3Type LieTensor:
     tensor([[ 0.1606,  0.0232, -1.5516],
             [-0.0807, -0.7184, -0.1102]], requires_grad=True)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,21 +87,25 @@ pytest
 ```python
 >>> import torch, pypose as pp
 
->>> # A random so(3) LieTensor
+>>> # Create a random so(3) LieTensor with size 2x3x3 and enable gradient tracking
```
wang-chen, line 90: so(3) is lie algebra, size is 2x3, not 2x3x3.

README.md
```diff
@@ -87,21 +87,25 @@ pytest
...
 >>> r = pp.randn_so3(2, requires_grad=True)
     so3Type LieTensor:
     tensor([[ 0.1606,  0.0232, -1.5516],
             [-0.0807, -0.7184, -0.1102]], requires_grad=True)
 
+    # Compute the matrix exponential of the so(3) LieTensor to get an SO(3) LieTensor
```
wang-chen, line 96: we use quaternion, not matrix.

wang-chen (changes requested):
Hi, @Aryan4884 thanks for your pull request, but there is incorrect information in your pull request. In my opinion, if the codes are self-explainable, we don't need to provide comments.  We only provide necessary comments for conciseness and readability.

Issue #302, opened 2023-10-29
mat2Sim3


Chensiyu00:
### 🐛 Describe the bug

**Report:**
> Exception has occurred: RuntimeError
> linalg.det: A must be batches of square matrices, but they are 3 by 4 matrices
>   File "/home/csy/pypose/evo_pypose/align_and_sync.py", line 316, in svdstf
>     return mat2Sim3(T, check=True)
>   File "/home/csy/pypose/evo_pypose/align_and_sync.py", line 368, in <module>
>     r1, t1, s1 = svdstf(est.translation().unsqueeze(0).repeat(10,1,1), gt.translation().unsqueeze(0).repeat(10,1,1), with_scale=True)
> RuntimeError: linalg.det: A must be batches of square matrices, but they are 3 by 4 matrices
> 

Step to reproduce:
Can not accept any matrix with 3*4 input 
> mat2Sim(T, check=True/false) T in [...,3,4]

The bug should from the line 394 in convert.py

391    shape = mat.shape
392    rot = mat[..., :3, :3]
393
394    s = torch.pow(torch.det(mat), 1/3).unsqueeze(-1)
should replace the mat with rot

@wang-chen 

### Versions

Collecting environment information...
PyTorch version: 2.0.0+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.6 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
Clang version: Could not collect
CMake version: version 3.26.4
Libc version: glibc-2.31

Python version: 3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-5.15.0-87-generic-x86_64-with-glibc2.31
Is CUDA available: True
CUDA runtime version: 11.7.64
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 4080
Nvidia driver version: 525.116.04
cuDNN version: Probably one of the following:
/usr/local/cuda-11.7/targets/x86_64-linux/lib/libcudnn.so.8.6.0
/usr/local/cuda-11.7/targets/x86_64-linux/lib/libcudnn_adv_infer.so.8.6.0
/usr/local/cuda-11.7/targets/x86_64-linux/lib/libcudnn_adv_train.so.8.6.0
[...]

PR #304, opened 2023-10-29
fix `mat2Sim3` bug #302

wang-chen:
To fix the bug reported in #302


wang-chen:
bug fixed, you may have a test now. @Chensiyu00

Chensiyu00:
fixed

PR #313, opened 2023-12-09
assert x does not contain nan in CG solver

hxu296:
Currently, CG will silently return `x` even if it contains `nan`, which may hide numeric instability issue.

### DIFF
--- a/pypose/optim/solver.py
+++ b/pypose/optim/solver.py
@@ -327,4 +327,7 @@ class CG(nn.Module):
             r -= alpha.unsqueeze(-1)*q
             rho_prev = rho_cur
 
+        assert not torch.any(torch.isnan(x)), \
+            'Conjugate Gradient Solver Failed. Check your matrix (may not be Symmetric Positive Definite)'
+
         return x


pypose/optim/solver.py
```diff
@@ -327,4 +327,7 @@ def forward(self, A: Tensor, b: Tensor, x: Optional[Tensor]=None,
             r -= alpha.unsqueeze(-1)*q
             rho_prev = rho_cur
 
+        assert not torch.any(torch.isnan(x)), \
+            'Conjugate Gradient Solver Failed. Check your matrix (may not be Symmetric Positive Definite)'
```
wang-chen, line 331: Line is too long, refer to point 2 in this [guideline](https://github.com/pypose/pypose/wiki/Developer-Guideline).

wang-chen (changes requested):

wang-chen:
You can also use this function to check: https://pypose.org/docs/main/generated/pypose.hasnan/?highlight=nan#pypose.hasnan

wang-chen (approved):

PR #315, opened 2023-12-11
Implementation of Radius Outlier Removal (ROR) for Point Cloud Processing

Murphy41:
Radius Outlier Removal (ROR) is a technique used in point cloud processing to identify and remove outlier points based on the number of neighboring points within a specified radius. It classifies points as outliers if they have fewer neighbors than a defined threshold within that radius, effectively filtering noise or sparse points from the data.

The current implementation outputs a boolean mask instead of directly manipulating the point cloud data. This approach is chosen to address challenges associated with varying tensor dimensions across different batches in batch-processed data. The mask indicates whether each point in the point cloud is considered an outlier (False) or not (True). Users can apply this mask to their data as needed.


Inoriros (approved):
Perfect to me. Just replaced words that may cause confusion.

wang-chen (approved):

PR #317, opened 2023-12-13
bsr/mm

ZihangFang:
support mm between bsr and bsc.


zitongzhan:
@chen-wang this is the bsr @ bsc operation as previously appeared in dev meeting. If you have time, could you please review the PR?

wang-chen (approved):

PR #318, opened 2023-12-17
luyr/geodesic_loss

luyr:
Add geodesic loss for rotation and pose loss for rotation+translation.


wang-chen:
need both [functions](https://pytorch.org/docs/stable/generated/torch.nn.functional.mse_loss.html) and [modules](https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html)

PR #325, opened 2023-12-30
Dynamic system revision

Lx017:


Lx017:
have slight issue with bmv and pnp test

PR #334, opened 2024-01-29
Implementation of  2 downsampling methods for Point Cloud - random_downsample; voxel_filter

YuqingLi-hub:
The downsampling method is used to unify data size and reduce computational cost. Here, we implement two commonly used methods.

Random_downsample:
This function takes two arguments: a point cloud and the desired number of points after downsampling. It outputs a randomly selected downsampled point cloud.

Voxel_filter:
This function conducts voxel filtering on a provided point cloud. It has three arguments: the given point cloud, voxel size in three dimensions, and an optional boolean value that indicates whether to randomly select a point from each voxel as a representative or use the mean of all points in each voxel. The output is the downsampled point cloud, where each point serves as the representative point in its corresponding voxel.


wang-chen:
I can see the status of this PR is still in progress, if you are ready for review, please let me know.

YuqingLi-hub:
> I can see the status of this PR is still in progress, if you are ready for review, please let me know.

Hi Wang Chen,
It's ready to review now.

pypose/function/geometry.py
```diff
@@ -356,3 +356,102 @@ def svdtf(source, target):
...
+    if random:
+        sorting_indices = torch.argsort(inverse_indices).squeeze()
+        sorted_points = points[sorting_indices, :]
+        random_indices = torch.cat([
+            torch.randint(low=0, high=count.item(), size=(1,)) for count in counts])
+        selected_indices = (random_indices + torch.cumsum(counts,dim=0) - counts).squeeze()
+        downsampled_points = sorted_points[..., selected_indices, :]
+        return downsampled_points
+    else:
+        voxel_means = torch.zeros_like(unique_indices,dtype=torch.float32)
```
pyposebot, line 449: `dtype=input.dtype`

pypose/function/geometry.py
```diff
@@ -356,3 +356,102 @@ def svdtf(source, target):
...
+        voxel_counts = torch.zeros(unique_indices.size(0), dtype=torch.float32)
+
+        voxel_means.index_add_(0, inverse_indices, points[..., :3])
+        voxel_counts.index_add_(0, inverse_indices,
+                                torch.ones_like(inverse_indices, dtype=torch.float32))
```
pyposebot, line 454:
```
ones = torch.ones_like(inverse_indices, dtype=input.dtype)
voxel_counts.index_add_(0, inverse_indices, ones)
```

pypose/function/geometry.py
```diff
@@ -356,3 +356,102 @@ def svdtf(source, target):
...
+    min_bound = torch.min(points[..., :3], dim=-2).values
+    voxel_indices = ((points[..., :3] - min_bound) / torch.tensor(leaf_size)).to(torch.int64)
+
+    unique_indices, inverse_indices,counts = torch.unique(
+        voxel_indices, dim=-2, return_inverse=True, return_counts=True)
+    if random:
+        sorting_indices = torch.argsort(inverse_indices).squeeze()
+        sorted_points = points[sorting_indices, :]
+        random_indices = torch.cat([
+            torch.randint(low=0, high=count.item(), size=(1,)) for count in counts])
```
pyposebot, line 444: define intermediate variables if the line is too long.

pyposebot (changes requested):

pypose/function/geometry.py
```diff
@@ -356,3 +356,102 @@ def svdtf(source, target):
...
+        tensor([[ 4.,  5.,  6.],
+                [ 1.,  2.,  3.],
+                [10., 11., 12.]])
+    '''
+    assert points.size(-1) >= 3, "The last dimension of the pointcloud should exceed 3."
+    assert num_points <= points.size(-2), "Number of points to downsample to must be \
+        less than or equal to the number of points in the point cloud."
+
+    indices = torch.randperm(points.size(-2))[:num_points]
+    downsampled_points = points[..., indices, :]
```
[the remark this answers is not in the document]
YuqingLi-hub: torch.index_select

wang-chen:
Hi, I renamed some names and tried to make these functions work for arbitrary dimensional space such 2D.
You also need to add auto tests for those functions, such as [this test](https://github.com/pypose/pypose/blob/main/tests/function/test_spline.py).

wang-chen:
Hi @YuqingLi-hub , Thank you so much for this update. The provided testing codes are to be working fine.
I tried the following test codes, it seems that they cannot run for **batched** tensors.

```python
points = torch.randn(2, 10, 3)
num = 5
result = pp.geometry.random_filter(points, num)
result = pp.geometry.voxel_filter(points, [1., 1., 1.])
```

For `random_filter`, I changed to the following code, it seems working.
```python
def random_filter(points, num):
    assert points.size(-1) >= 1, "The last dim of the points should not less than 1."
    assert num <= points.size(-2), "Number of points to sample must not larger than " \
        "the number of input points."
    
    indices = torch.randperm(points.size(-2))[:num]
    return points[..., indices, :]
```
I am not sure how to change `voxel_filter`, as we may need to change multiple places to make it to be compatible with batched operations.

wang-chen:
Until now, I have tried:
Change  [this line](https://github.com/pypose/pypose/blob/46cc5a6dae05d69dc14e1db321dc588f54f4c695/pypose/function/geometry.py#L436) to
```python
minp = torch.min(points[..., :dim], dim=-2, keepdim=True).values
```
[this line](https://github.com/pypose/pypose/blob/46cc5a6dae05d69dc14e1db321dc588f54f4c695/pypose/function/geometry.py#L450) to
```python
voxel_counts = torch.zeros(unique_indices.shape[:-1], dtype=points.dtype)
```
You may want to access a specific dimension with negative numbers, as we may have multiple batches, e.g., points can be a shape of `[... ,5,3]`, where `...` can be any dimensions.

wang-chen (approved):

PR #352, opened 2024-09-11
Batched UFK and PF

FuTaimeng:
The UKF and PF have been updated to support batched operations, allowing for a batch dimension in states, covariances, and inputs. Internally, the sigma points/particles dimension is now placed before the batch dimension (if present) for computational purposes. However, this does not affect the user interface, where all exposed variables either have the batch dimension as the first dimension or no batch dimension if not needed.

Additionally, an option has been added in UKF for performing positive-definite (PD) correction on the covariance matrix P. When enabled, this projects P onto the PD space by clamping small eigenvalues to a user-specified eps value. This feature improves algorithm stability in challenging scenarios.

### DIFF
--- a/pypose/module/pf.py
+++ b/pypose/module/pf.py
@@ -94,26 +94,28 @@ class PF(EKF):
 
         3. Prepare data
 
-        >>> T, N = 5, 2 # steps, state dim
-        >>> states = torch.zeros(T, N)
-        >>> inputs = torch.randn(T, N)
-        >>> observ = torch.zeros(T, N)
+        >>> B, T, N = 3, 5, 2 # batchsize, steps, state dim
+        >>> states = torch.zeros(B, T, N)
+        >>> inputs = torch.randn(B, T, N)
+        >>> observ = torch.zeros(B, T, N)
         >>> # std of transition, observation, and estimation
         >>> q, r, p = 0.1, 0.1, 10
-        >>> Q = torch.eye(N) * q**2
-        >>> R = torch.eye(N) * r**2
-        >>> P = torch.eye(N).repeat(T, 1, 1) * p**2
-        >>> estim = torch.randn(T, N) * p
+        >>> Q = torch.eye(N).repeat(B, 1, 1) * q**2
+        >>> R = torch.eye(N).repeat(B, 1, 1) * r**2
+        >>> P = torch.eye(N).repeat(B, T, 1, 1) * p**2
+        >>> estim = torch.randn(B, T, N) * p
 
         4. Perform PF prediction. Note that estimation error becomes smaller with more steps.
 
         >>> for i in range(T - 1):
-        ...     w = q * torch.randn(N) # transition noise
-        ...     v = r * torch.randn(N) # observation noise
-        ...     states[i+1], observ[i] = model(states[i] + w, inputs[i])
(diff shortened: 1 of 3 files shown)


pypose/module/pf.py
```diff
@@ -187,10 +189,10 @@ def resample_particles(self, q, x):
         r'''
         Resample the set of a posteriori particles
         '''
-        r = torch.rand(self.particles, dtype=x.dtype, device=x.device)
-        cumsumq = torch.cumsum(q, dim=-1)
-        return x[torch.searchsorted(cumsumq, r)]
+        r = torch.rand(x.shape[:-1], dtype=x.dtype, device=x.device)
+        index = torch.searchsorted(torch.cumsum(q, dim=0), r)
+        return x[index, torch.arange(x.shape[1]).unsqueeze(0)] if x.dim()==3 else x[index]
```
wang-chen, line 194: can we implement this line using `torch.index_select` or `torch.gather` without using `if else`?

pypose/module/ukf.py
```diff
@@ -156,9 +163,10 @@ class UKF(EKF):
           Cleveland State University, 2006
     '''
 
-    def __init__(self, model, Q=None, R=None, msqrt=None):
+    def __init__(self, model, Q=None, R=None, msqrt=None, correct_PD_eps=-1):
```
wang-chen, line 166: How about changing the interface to 'makepd = False'?

pypose/module/ukf.py
```diff
@@ -187,20 +195,24 @@ def forward(self, x, y, u, P, Q=None, R=None, t=None, k=None):
...
+        ye = (w * ys).sum(dim=0)
         ey = ye - ys
         Py = self.compute_cov(ey, ey, w, R)
 
         Pxy = self.compute_cov(ex, ey, w)
         K = Pxy @ pinv(Py)
         x = xe + bmv(K, y - ye)
         P = P - K @ Py @ K.mT
+        if self.correct_PD_eps >= 0:
+            P = self.project_to_positive_definite(P)
```
wang-chen, line 215: We can define a new global function like `clamp2pd` to implement 'Positive Definite Eigenvalue Clamping' method so that other algorithms can also use it.

pypose/module/ukf.py
```diff
@@ -217,15 +229,21 @@ def sigma_weight_points(self, x, P, k):
...
         '''Compute covariance of two set of variables.'''
         a, b = a.unsqueeze(-1), b.unsqueeze(-1)
-        return Q + (w.unsqueeze(-1) * a @ b.mT).sum(dim=-3)
+        return Q + (w.unsqueeze(-1) * a @ b.mT).sum(dim=0)
+
+    def project_to_positive_definite(self, A):
+        '''Project matrix to positive definite by clamping its eigenvalues to positive.'''
+        eigenvalues, eigenvectors = torch.linalg.eigh(A)
+        eigenvalues = torch.clamp(eigenvalues, min=self.correct_PD_eps)
+        return (eigenvectors @ torch.diag_embed(eigenvalues)) @ eigenvectors.mT
```
wang-chen, line 249: how about define the global function.

wang-chen (changes requested):

wang-chen:
need to check batched ekf, ukf, and pf, as well as their tests.

Issue #356, opened 2024-12-10
`pp.LieTensor` does not annotate the abstract methods properly


MarkChenYutian:
## Summary

The type annotation for `pp.LieTensor` can be improved for downstream projects to use static type checkers like `PyRight` or `mypy`.

### Current Problem

Currently, due to the lack of type annotation, the type of `LieTensor.Act` is inferred as `Unknown -> NoReturn` in VSCode Pyright extension.

![Screenshot 2024-12-10 at 2 03 01 AM](https://github.com/user-attachments/assets/7f4aef55-fb5a-4818-b810-bb39f3c5d77b)

This causes the type analyzer to falsely report that all code below `LieTensor.Act` is not reachable.

![Screenshot 2024-12-10 at 2 03 45 AM](https://github.com/user-attachments/assets/3ab1c76d-25b2-45fb-b512-ea84b4c4c0ae)

The reason for this undesired behavior is discussed in [this comment of PyRight community](https://github.com/microsoft/pylance-release/issues/248#issuecomment-675523999). Since `class LieType` in `pypose/lietensor/lietneor.py` did not annotate the `Act` method as an abstract method (overridden by specific Lie-type implementation such as `SE3` or `SO3`), the type checker believes `LieType.Act` will raise exception regardless of the input according to the following code snippet:

https://github.com/pypose/pypose/blob/62a1d13c15fb56f4f68d17e6d6f01a8850bc545a/pypose/lietensor/lietensor.py#L77-L81

In reality, this will not happen since `LieType` is only an abstract base class and we will always use some sub-class of `LieType` which contains the manifold-specific implementation.

## Improvements

By properly annotating the type, the downstream project can leverage static analysis tools to identify many potential bugs in the project without running slow unit tests. For instance, in the [MAC-VO](https://github.com/MAC-VO/MAC-VO) project, we integrated PyRight static type checking as part of the GitHub Action pipeline.

In general, a better type annotation allows downstream developers to develop faster and safer with static analysis tools and IDE support.

## Risks

Since the type annotation of Python in general does not change the behavior of runtime, there should be minimal risk in the proposed refactor. Marking the `LieType` class as an abstract base class (`abc.ABC`) will indeed change the behavior of class. Specifically, the class cannot be instantiated anymore. However, I don't think the current PyPose codebase ever instantiated a `LieType` instance. We only use the subclasses of `LieType` such as `SE3Type`.

## Involved components

* `pypose/lietensor/lietensor.py`

[...]

PR #367, opened 2025-01-17
Improve typing annotation for LieTensor and LieType.

MarkChenYutian:
Fix issue 
* #356 

Originally I planned to convert all methods of the `class LieType` into abstract methods. However, this is not feasible as each specific lie type (SE3, se3, for instance) only implements a subset of methods implemented in the `LieType` (depending on it is a Lie Group or Lie Algebra).

Instead, I manually annotate the type annotations for the base methods in `class LieType` so the static type checker will no longer reduce the type to `-> NoReturn` automatically.

### DIFF
--- a/pypose/lietensor/lietensor.py
+++ b/pypose/lietensor/lietensor.py
@@ -56,35 +56,35 @@ class LieType(ABC):
     def on_manifold(self) -> bool:
         return self.dimension == self.manifold
 
-    def add_(self, input: LieTensor, other: LieTensor) -> torch.Tensor:
+    def add_(self, input: LieTensor, other: torch.Tensor) -> torch.Tensor:
         if self.on_manifold:
             other1 = torch.Tensor.as_subclass(input, torch.Tensor)
             other2 = torch.Tensor.as_subclass(other, torch.Tensor)
             return input.copy_(other1 + other2[..., :self.manifold[0]])
         raise NotImplementedError("Instance has no add_ attribute.")
 
-    def Log(self, X: LieTensor | torch.Tensor) -> LieTensor:
+    def Log(self, X: LieTensor) -> LieTensor:
         if self.on_manifold:
             raise AttributeError("Lie Algebra has no Log attribute")
         raise NotImplementedError("Instance has no Log attribute.")
 
-    def Exp(self, x: LieTensor | torch.Tensor) -> LieTensor:
+    def Exp(self, x: LieTensor) -> LieTensor:
         if not self.on_manifold:
             raise AttributeError("Lie Group has no Exp attribute")
         raise NotImplementedError("Instance has no Exp attribute.")
 
-    def Inv(self, X: LieTensor | torch.Tensor) -> LieTensor:
+    def Inv(self, X: LieTensor) -> LieTensor:
         if self.on_manifold:
             return LieTensor(-X, ltype=self)
         raise NotImplementedError("Instance has no Inv attribute.")
 
-    def Act(self, X: LieTensor | torch.Tensor, p: torch.Tensor) -> torch.Tensor:
+    def Act(self, X: LieTensor, p: torch.Tensor) -> torch.Tensor:
         """ action on a points tensor(*, 3[4]) (homogeneous)"""
         if not self.on_manifold:
             raise AttributeError("Lie Group has no Act attribute")
         raise NotImplementedError("Instance has no Act attribute.")
 
-    def Mul(self, X: LieTensor | torch.Tensor, Y: LieTensor | torch.Tensor) -> LieTensor:
+    def Mul(self, X: LieTensor, Y: numbers.Number | torch.Tensor | LieTensor) -> LieTensor:
         if not self.on_manifold:
             raise AttributeError("Lie Group has no Mul attribute")
         raise NotImplementedError("Instance has no Mul attribute.")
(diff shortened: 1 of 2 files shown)


pypose/lietensor/lietensor.py
```diff
@@ -319,13 +334,13 @@ def matrix(self, input):
...
+    def Jr(self, X):
         """
         Right jacobian of so(3)
         """
-        K = vec2skew(x)
-        theta = torch.linalg.norm(x, dim=-1, keepdim=True).unsqueeze(-1)
-        I = torch.eye(3, device=x.device, dtype=x.dtype).expand(x.lshape+(3, 3))
+        K = vec2skew(X)
+        theta = torch.linalg.norm(X, dim=-1, keepdim=True).unsqueeze(-1)
+        I = torch.eye(3, device=X.device, dtype=X.dtype).expand(X.lshape+(3, 3))
```
wang-chen, line 343: We use `x` to denote Lie algebra, `X` for Lie group.

pypose/lietensor/lietensor.py
```diff
@@ -29,100 +33,103 @@
...
-    def add_(self, input, other):
+    def add_(self, input: LieTensor, other: LieTensor) -> torch.Tensor:
         if self.on_manifold:
             other1 = torch.Tensor.as_subclass(input, torch.Tensor)
             other2 = torch.Tensor.as_subclass(other, torch.Tensor)
             return input.copy_(other1 + other2[..., :self.manifold[0]])
         raise NotImplementedError("Instance has no add_ attribute.")
 
-    def Log(self, X):
+    def Log(self, X: LieTensor | torch.Tensor) -> LieTensor:
```
wang-chen, line 66: I see a lot of `LieTensor | torch.Tensor`, but I think not all functions allow `torch.Tensor` as input. For example, I remember this `Log` function may be one of them, it only allows Lie group LieTensor.

wang-chen (changes requested):
I think this PR has very high quality. I only have some minor comments. We are ready to merge once they are addressed.

pypose/lietensor/lietensor.py
MarkChenYutian: I refactored the argument name because by standard, the overriding method should have exactly the same function signature as the base method.


Consider this situation: If the base method uses `X`

```python
class LieType:
    def Jr(self, X) -> LieTensor: ...
```

and the overriding method has a different parameter name `X -> x`.

```python
class SE3_Type(LieType):
    def Jr(self, x) -> LieTensor: ...
```

Then when someone uses a keyword argument (say `some_type.Jr(X= ...)`), and the `some_type` is an instance of `SE3_Type`, an unexpected exception will be raised (as there is no argument with name `X` in `SE3_Type.Jr`)

pypose/lietensor/lietensor.py
wang-chen: A line like `x = x.tensor() if isinstance(x, LieTensor) else x` is to tolerate unintentional inputs, and even if that happens, the program doesn't raise an error without telling any cues. But as a typing annotation, we should not tell the users it is designed to accept `Tensor`.
 However, I believe some functions are accepting both `Tensor` and `LieTensor`, we need to be careful about that.

pypose/lietensor/lietensor.py
wang-chen: This explanation is acceptable. Thank you.

MarkChenYutian:
I've corrected the type annotation to follow the documentation provided on `pypose.org`.

pypose/lietensor/lietensor.py
```diff
@@ -29,100 +33,103 @@
...
+    def manifold(self) -> torch.Size:
         return self._manifold
 
     @property
-    def on_manifold(self):
+    def on_manifold(self) -> bool:
         return self.dimension == self.manifold
 
-    def add_(self, input, other):
+    def add_(self, input: LieTensor, other: torch.Tensor) -> torch.Tensor:
```
wang-chen, line 59: I believe `add_` returns `LieTensor` as [doc](https://pypose.org/docs/main/generated/pypose.add/#pypose.add). This one is used internally for updating LieTensor by gradients via various optimizers.

pypose/lietensor/lietensor.py
```diff
@@ -29,100 +33,103 @@
...
 
-    def Act(self, X, p):
+    def Act(self, X: LieTensor, p: torch.Tensor) -> torch.Tensor:
         """ action on a points tensor(*, 3[4]) (homogeneous)"""
         if not self.on_manifold:
             raise AttributeError("Lie Group has no Act attribute")
         raise NotImplementedError("Instance has no Act attribute.")
 
-    def Mul(self, X, Y):
+    def Mul(self, X: LieTensor, Y: numbers.Number | torch.Tensor | LieTensor) -> LieTensor:
```
wang-chen, line 87: `Mul` returns `Tensor` or `LieTensor` as [doc](https://pypose.org/docs/main/generated/pypose.mul/?highlight=mul)

wang-chen (changes requested):

MarkChenYutian:
Sorry I forgot to check the return type, only focused on the parameter type in previous commit. I will check the return types now.

MarkChenYutian:
I've checked through the return type of all methods in `class LieType` with doc on pypose.org, should be good now

wang-chen (approved):

Issue #344, opened 2024-05-24
MPC examples do not work


FenglongSong:
### 🐛 Describe the bug

Hi,

Thanks for developing this great repo. I'm interested in using the MPC modules of pypose. When I tried to run both examples in `examples/mpc/` folder, I got some errors.

1. This is what I got when running `examples/module/mpc/cartpole.py`:
```
(leggedrobot) ➜  pypose git:(main) ✗ python3 examples/module/mpc/cartpole.py
Namespace(device='cpu', save='./examples/module/mpc/save/', show=False)
Traceback (most recent call last):
  File "/Users/fenglongsong/Documents/GitHub/pypose/examples/module/mpc/cartpole.py", line 104, in <module>
    traj_loss.backward()
  File "/Users/fenglongsong/miniconda3/envs/leggedrobot/lib/python3.10/site-packages/torch/_tensor.py", line 522, in backward
    torch.autograd.backward(
  File "/Users/fenglongsong/miniconda3/envs/leggedrobot/lib/python3.10/site-packages/torch/autograd/__init__.py", line 266, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Output 3 of UnbindBackward0 is a view and its base or another view of its base has been modified inplace. This view is the output of a function that returns multiple views. Such functions do not allow the output views to be modified inplace. You should replace the inplace operation by an out-of-place one.
```

2. This is what I got when running `examples/module/mpc/linear.py`:
```
(leggedrobot) ➜  pypose git:(main) ✗ python3 examples/module/mpc/linear.py
Namespace(device='cpu', save='./examples/module/mpc/save/', show=False)
Traceback (most recent call last):
  File "/Users/fenglongsong/Documents/GitHub/pypose/examples/module/mpc/linear.py", line 81, in <module>
    traj_loss.backward()
  File "/Users/fenglongsong/miniconda3/envs/leggedrobot/lib/python3.10/site-packages/torch/_tensor.py", line 522, in backward
    torch.autograd.backward(
  File "/Users/fenglongsong/miniconda3/envs/leggedrobot/lib/python3.10/site-packages/torch/autograd/__init__.py", line 266, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[...]

PR #369, opened 2025-01-26
Allow differentiation on MPC

andreaostuni:
this pull request resolves the issues for allowing differentiation in a batched manner through the MPC layer resolving the issues emerged in #344

### DIFF
--- a/examples/module/dynamics/cartpole.py
+++ b/examples/module/dynamics/cartpole.py
@@ -14,24 +14,47 @@ class CartPole(pp.module.NLS):
         self.polemassLength = self.polemass * self.length
         self.totalMass = self.cartmass + self.polemass
 
-    def state_transition(self, state, input, t = None):
-        x, xDot, theta, thetaDot = state
-        force = input.squeeze()
-        costheta = theta.cos()
-        sintheta = theta.sin()
+    def state_transition(self, state, input, t=None):
+        """
+        Vectorized state transition function for batch operations.
 
-        temp = (force + self.polemassLength * thetaDot**2 * sintheta) / self.totalMass
+        Parameters:
+            state (torch.Tensor): Tensor of shape (batch_size, 4)
+            representing the state [x, xDot, theta, thetaDot].
+            input (torch.Tensor): Tensor of shape (batch_size, 1)
+              representing the input force.
+            t (torch.Tensor or None): Optional time variable
+            (not used in this implementation).
 
-        thetaAcc = (self.gravity * sintheta - costheta * temp) / \
-            (self.length * (4.0 / 3.0 - self.polemass * costheta**2 / self.totalMass))
-    
+        Returns:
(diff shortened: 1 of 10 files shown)


rokeeeto-li:
Hi Andrea,

Thank you for your pull request and for your contributions to PyPose. We truly appreciate the time and effort you've put into this.

As maintainers of the library, we take a cautious approach when updating core functions, so we are unable to merge your PR directly. However, we would love to discuss potential refinements with you to further optimize the library.

In general, we aim to follow these principles for pull requests:
1. Address unrelated issues through separate PRs.
2. Keep changes as minimal as possible within a single PR.
3. When making modifications, prioritize resolving issues at a higher level rather than altering deeper-level functions unless necessary.

Based on these principles, I will provide feedback on your PR. We encourage you to submit a revised PR incorporating these suggestions. If you prefer, we are also happy to make the necessary adjustments ourselves. Either way, we greatly appreciate your contributions and will ensure you receive due credit for your efforts.

In your PR, I see that you primarily addressed two issues:

1. Fixing an in-place operation bug in the MPC example that affects gradient backpropagation.
2. Resolving an issue with how the current implementation of dynamics computes the Jacobian (A, B) for batched data.

To facilitate review and merging, I believe it would be best to submit these fixes as two separate PRs. Below are my thoughts on each issue:

**Issue 1: MPC Example Bug** 
The two MPC examples currently encounter an error due to a mismatch between the state version recorded during the state transition and the one used during backpropagation. One potential fix is to clone the state at the beginning of the state transition, ensuring that each state is assigned a separate computational graph node.
Specifically, we propose modifying the following lines:

- In example/module/mpc/cartpole.py, update line 22: `x, xDot, theta, thetaDot = state.squeeze()`  to `x, xDot, theta, thetaDot = state.clone().squeeze() `
- In pypose/module/dynamics.py, update line 213:  `z = bmv(self.A, state) + bmv(self.B, input)` to `z = bmv(self.A, state.clone()) + bmv(self.B, input)`  

We believe this should resolve the issue. However, if you have a better solution, we would love to discuss it further.

**Issue 2: Jacobian Shape in Dynamics** 
Currently, dynamics.A/B uses torch.autograd.functional.jacobian, which, when computing the Jacobian for batched inputs (shape [b, m] w.r.t. [b, n]), results in an unexpected output shape of [b, m, b, n] instead of [b, m, n]. Ideally, this modification should be made in dynamics.py.

I had previously noticed this issue and used a similar workaround in my private version. However, this approach remains debatable—it introduces an additional Jacobian computation along the batch dimension, increasing computational cost. Moreover, from a coding perspective, it is not the most elegant solution. A more optimal approach would involve using jacrev combined with vmap to achieve efficient parallel computation. Due to these concerns, the fix for this bug is still under consideration.


Once again, thank you for your contributions to PyPose! Additionally, in line with our second guideline, we kindly ask that you minimize unrelated modifications when submitting code, such as avoiding automatic formatting changes (e.g., altering quotation marks).

PR #381, opened 2025-11-20
Seokwoopark/bsr matrix multiplication implementation

SEOKWOOPARK:
Currently, there is no function for BSR (Block Sparse Row) multiplication in pypose. So, I implemented it using Triton. Descriptions for each function are as follows. These functions have been written in pypose > pypose > sparse > bsr_mm.py.

- **bsr_mm_count_coeffs**: This Triton kernel runs one program per x block-row and counts how many candidate z blocks can be generated from that row by looking up the referenced y rows. It writes per-x-block candidate counts into block_counts shifted by +1 for prefix-sum layout and decides whether the row should be represented in list-mode or compact range-mode via row_min. It also represents z_nnz into block_counts[0] so the later prefix-sum can start from an existing offset.

- **bsr_mm_list_coeffs**: This Triton kernel takes the prefix-summed mm_offsets and maps each flattened candidate id (mm_block) to a concrete (row, col) pair for z. It binary-searches mm_offsets to find which x block produced the candidate and binary-searches x_offsets to recover the owning x block-row. It emits candidate (mm_rows, mm_cols) and records mm_src_blocks when list-mode is used like range-mode emits src = -1.

- **bsr_mm_numeric_kernel**: This Triton kernel performs the actual numeric block multiplication for z = x @ y using the finalized z structure (z_offsets, z_columns). Each program computes one z block-row and iterates over that row’s output blocks. The function also accumulates dense block products from matching intermediate k indices. It writes the resulting dense blocks directly into z_values_ptr.

- **bsr_mm_triton**: This host function orchestrates the full pipeline: count candidates (bsr_mm_count_coeffs) → prefix-sum → list candidates (bsr_mm_list_coeffs) → unique/sort to finalize z structure → numeric compute (bsr_mm_numeric_kernel). It allocates all intermediate buffers, converts the candidate list into final (z_offsets, z_columns), and launches the numeric kernel to fill z_vals. It returns the final BSR output (z_offsets, z_columns, z_vals).

- **bsr_output_to_dense_numpy**: This reconstructs a full dense matrix from the BSR output.

The corresponding test code has been added to pypose > test > sparse > test_bsr.py. To test multiple cases, I repeated the test flow 20 times with randomly generated inputs, referring to the existing codebase in test_bsr.py. My test code compares the Triton-based result with PyTorch dense matrix multiplication (‘@’).

After several trials for implementing, it was refactored so that the whole structure including variables' names and functions' name follows Nvidia-warp([URL](https://github.com/NVIDIA/warp/blob/964575873c234735a1fd4a40d31b96dc436f7d09/warp/_src/sparse.py#L1739), warp/warp/_src/sparse.py) -> bsr_mm_count_coeffs, bsr_mm_list_coeffs


zitongzhan:
@SEOKWOOPARK Please read the doc of triton. As discussed, there's ENV variable that allows debugging triton kernel on CPU. Ideally this flag should be used to test triton implementations inside CPU-only CI runtime.

zitongzhan:
@gemini-code-assist This implementation is supposed to be a mirror of Warp's BSR mm kernels. Could you check if it follows the design of https://github.com/NVIDIA/warp/blob/v1.6.0/warp/sparse.py#L1537-L1771

tests/sparse/test_bsr.py
```diff
@@ -1,6 +1,14 @@
 import torch
 import pytest
 import pypose as pp
+from pypose.sparse import bsr_mm_triton, bsr_output_to_dense_numpy
+
+try:
+    import triton
```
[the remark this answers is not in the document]
SEOKWOOPARK: You're correct. I tried it just for fixing an error. But, it's being utilized in the current state

pypose/sparse/bsr_mm.py
```diff
@@ -0,0 +1,318 @@
...
+
+    A_num_block_rows: number of block-rows of A and C
+    A_num_block_cols: number of block-columns of A and block-rows of B
+    B_num_block_cols: number of block-columns of B and C (kept for completeness. C’s pattern is derived from A/B offsets)
+
+    Returns:
+    C_offsets: 1D int tensor, crow_indices of C, shape [A_num_block_rows + 1]
+    C_columns: 1D int tensor, col_indices of C, shape [C_all_block_nums]
+    C_vals: 3D tensor, shape [C_all_block_nums, A_block_row_num, B_block_col_num]. Block values of the result C in BSR format.
+    '''
```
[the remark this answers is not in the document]
zitongzhan: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/pypose/pypose/pull/381#discussion_r2554588981)

pypose/sparse/bsr_mm.py
```diff
@@ -0,0 +1,293 @@
+import torch
+import triton
+import triton.language as tl
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
```
zitongzhan, line 5: `device = "cuda" if torch.cuda.is_available() else "cpu"` should not appear under global scope. Check other files in the codebase, e.g., epnp, to see how device is determined.

pypose/sparse/bsr_mm.py
```diff
@@ -0,0 +1,455 @@
...
+    *,
+    COUNT_BLOCK: int = 128,
+    LIST_NUM_WARPS: int = 4,
+):
+    x_nrow = x_num_block_rows
+    x_nnz = int(x_offsets[-1].item())
+    y_ncol = y_num_block_cols
+    z_nnz = 0
+    copied_z_nnz = 0
+    row_min = torch.empty((x_nrow,), device=device, dtype=torch.int32)
```
SEOKWOOPARK: device = x_vals.device

pypose/sparse/bsr_mm.py
```diff
@@ -0,0 +1,456 @@
...
+        x_block_col_num_pow2=max(16, triton.next_power_of_2(x_bc)),
+        y_block_col_num_pow2=max(16, triton.next_power_of_2(y_bc)),
+        z_block_row_num=x_nrow,
+        DTYPE=tl.float32,
+    )
+
+    return z_offsets, z_columns, z_vals
+
+
+def bsr_output_to_dense_numpy(
```
zitongzhan: This function isn't necessary because `torch` already implements the `bsr.to_dense()` function.

zitongzhan (changes requested):

pypose/sparse/bsr_mm.py
```diff
@@ -0,0 +1,399 @@
...
+    z_block_row_indices = tl.arange(0, x_block_row_num_pow2)
+    z_block_col_indices = tl.arange(0, y_block_col_num_pow2)
+    z_block_offsets = z_block_row_indices[:, None] * y_block_col_num + z_block_col_indices[None, :]
+    z_mask = (z_block_row_indices[:, None] < x_block_row_num) & (z_block_col_indices[None, :] < y_block_col_num)
+
+    x_block_size = x_block_row_num * x_block_col_num
+    y_block_size = x_block_col_num * y_block_col_num
+    z_block_size = x_block_row_num * y_block_col_num
+
+    for z_idx in range(z_row_start_idx, z_row_end_idx):
```
zitongzhan, line 291: Why there are for loops used here? I think this is supposed to be parallelized in Triton SIMT kernel call, instead of some loops inside the kernel.

SEOKWOOPARK: Because I focused on correctness and implemented the formula as-is, I used nested loops. You're right about this. It should be done based on SIMD. However, we should consider two possible alternatives for parallelized SIMT:

1) **Reducing the loop (3 nested loops → 2 nested loops = a row-range for-loop + binary search with a constant 32-range for-loop)** is robust to sparse and non-uniform data and light on memory. And it is also safe at any size.

2) **Removing the loop (GEMM via dot product)** is consistent for scheduling and coalesced memory access.

How do you think about those 2 ways?

PR #406, opened 2026-08-09
Update PGO with BAE sparse solver

wang-chen:
Enhance the PGO functionality by integrating bae sparse solver, improving performance and efficiency in processing.

### DIFF
--- a/examples/module/pgo/readme.md
+++ b/examples/module/pgo/readme.md
@@ -1,11 +1,16 @@
 # Pose Graph Optimization
 
-An example for Pose Graph Optimization (PGO)
+An example of pose graph optimization (PGO) using either sparse or dense
+Levenberg-Marquardt optimization.
 
-## Installation
+## Requirements
 
-    python -m pip install -U matplotlib
-    pip install torchvision
+```bash
+python -m pip install -U matplotlib
+```
+
+Sparse optimization is enabled by default and requires CUDA. Use
+`--no-sparse --device cpu` to run the dense implementation on CPU.
 
 ## Prepare Dataset (Optional)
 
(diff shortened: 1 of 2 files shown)


examples/module/pgo/readme.md
```diff
@@ -41,9 +60,16 @@ Note: change `DATAROOT` to the folder you select.
...
           --vectorize          to accelerate computation
 
-## Note
+## Notes
 
-The current implementation of PGO is not using sparse matrices. Therefore, when the number of poses is very big, the memory consumption can be large (`--vectorize`) or running speed can be slow (`--no-vectorize`). The sparse matrices will be integrated in the next main release, while the API will be mostly unchanged, only internal logic will be updated.
+- Sparse mode is the default and is recommended for large pose graphs.
+- Dense mode incorporates the dataset information matrices but can require
+  considerably more memory.
+- `--vectorize` is enabled by default. Use `--no-vectorize` to reduce peak
```
zitongzhan: `--vectorize` is no longer needed because Sparse Jacobian is the new default
Sparse Jacobian doesn't have the concept of ``--no-vectorize`` because it is always vectorizing.

zitongzhan (changes requested):

examples/module/pgo/readme.md
wang-chen: `--vectorize` is used when `--sparse` is `False`.

zitongzhan: @copilot add an assert that `--no-vectorize` and `--sparse` are never used at the same time. Update docs `--vectorize` can be used when `--sparse` is False.

zitongzhan (approved):

wang-chen:
@zitongzhan After a consideration, totally remove full jacobian implementation from PGO, as it should be deprecated, and we should not tell users how to use the previous version.

zitongzhan:
> @zitongzhan After a consideration, totally remove full jacobian implementation from PGO, as it should be deprecated, and we should not tell users how to use the previous version.

Yes. That should cause less confusion.

